PR #4, opened 2017-08-02
Pull/build system fixes

Lyude:
These are numerous fixes for the build system that make it more consistent with most of the autotools build systems out there, along with fixing some obscure bugs. Additionally, we now just build the protocol marshalling code ourselves instead of bundling it.


Lyude:
(ignore the order of the commits that github lists here. I guess github makes the mistake of sorting commits by their creation date instead of the actual order they were committed to the repository, the spaces fix is actually the first commit in this pull request)

mvicomoya:
I manually rebased and merged these. I dropped the indentation commit, and replaced TABs by spaces in the subsequent patches where possible.

PR #7, opened 2017-09-11
More build system improvements

Lyude:
Since I finally managed to get Xwayland working with eglstreams, I figured it's about time to start cleaning things up and preparing to get these patches submitted at some point.
That being said, this patch series allows Xwayland and other projects to actually build against egl-wayland using pkg-config, so that we don't have to reinclude the same wayland protocol files for every project.

Wayland protocols are installed (by default) in `/usr/local/share/egl-wayland/wayland-eglstream*.xml`, similarly to the path scheme of the official wayland-protocols package. Everything else is installed in the usual directories.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,13 +1,13 @@
 AC_PREREQ([2.64])
 
-m4_define([wayland_external_major_version], [1])
-m4_define([wayland_external_minor_version], [0])
-m4_define([wayland_external_micro_version], [2])
-m4_define([wayland_external_version],
-          [wayland_external_major_version.wayland_external_minor_version.wayland_external_micro_version])
+m4_define([wayland_eglstream_major_version], [1])
+m4_define([wayland_eglstream_minor_version], [0])
+m4_define([wayland_eglstream_micro_version], [2])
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -6,7 +6,7 @@ m4_define([wayland_external_micro_version], [2])
 m4_define([wayland_external_version],
           [wayland_external_major_version.wayland_external_minor_version.wayland_external_micro_version])
 
-AC_INIT([wayland_external],
+AC_INIT([egl-wayland],
```
mvicomoya: Thank you very much for these patches.

I agree 'wayland_external' isn't a great name for this package. The 'external' bit of it was intended to relate to the [EGL External Platform interface](https://github.com/NVIDIA/eglexternalplatform) this library actually implements.

Also, I now think 'egl-wayland' wasn't the most convenient name for this project either (sorry for the bad naming choices) as it is actually an EGLStream-specific implementation.

I'm thinking that maybe we should change it to something along the lines of 'wayland-eglstream' instead.

Also, I think it'd be fine to search & replace all occurrences of 'wayland_external' by the new chosen name throughout this file.

PR #24, opened 2019-11-27
Add eglmesaext.h include in wayland-eglsurface.c

B4dM4n:
`eglmesaext.h` was previously included via `eglext.h` but mesa stopped providing GLVND headers in [mesa-19.2.2][1].
Without `eglmesaext.h` EGL_WAYLAND_Y_INVERTED_WL is not defined and wayland-eglsurface.c` fails to compile.

```log
../src/wayland-eglsurface.c: In function 'wlEglQueryNativeResourceHook':
../src/wayland-eglsurface.c:1523:10: error: 'EGL_WAYLAND_Y_INVERTED_WL' undeclared (first use in this function); did you mean 'EGL_WAYLAND_EGLSTREAM_WL'?
     case EGL_WAYLAND_Y_INVERTED_WL:
          ^~~~~~~~~~~~~~~~~~~~~~~~~
          EGL_WAYLAND_EGLSTREAM_WL
../src/wayland-eglsurface.c:1523:10: note: each undeclared identifier is reported only once for each function it appears in
```

[1]: https://gitlab.freedesktop.org/mesa/mesa/commit/8355658fa857536d948773b361c5ede770e637a3

### DIFF
--- a/src/wayland-eglsurface.c
+++ b/src/wayland-eglsurface.c
@@ -37,6 +37,7 @@
 #include <sys/types.h>
 #include <netinet/in.h>
 #include <fcntl.h>
+#include <EGL/eglmesaext.h>
 
 #define WL_EGL_WINDOW_DESTROY_CALLBACK_SINCE 3
 


mvicomoya:
@B4dM4n thanks for the PR! Rather including `EGL/eglmesaext.h`, I suggest adding `EGL_WAYLAND_Y_INVERTED_WL` to https://github.com/NVIDIA/egl-wayland/blob/master/wayland-egl/wayland-egl-ext.h so that we don't create a dependency for `EGL/eglmesaext.h`.

I can make that change if you cannot make amends to this PR.

mvicomoya:
@ekurzinger you might want to follow-up on this :-)

tpgxyz:
I've prepared RP #26 
egl-wayland compiles and runs fine on OpenMandriva Linux

erik-kz:
This fix should no longer be needed as of https://github.com/NVIDIA/egl-wayland/commit/4a343a4b563e84c6258efbddf1d910f9dd6e0300, which adds the EGL_WL_bind_wayland_display extension block to wayland-egl-ext.h, providing the missing enum definition.

PR #29, opened 2020-10-06
Install arch-specific pkgconfig file to libdir

kepstin:
The pkgconfig file wayland-eglstream.pc has instructions for linking to
an architecture specific library, so it should be installed to libdir
instead of datadir.

This is important on systems with multilib or multiarch, to make sure
that you link to the correct architecture version of the installed
libraries when multiple architectures are present.


erik-kz:
Sorry for the slow response. I believe you are correct, thank you very much for the patch.

Issue #34, opened 2021-02-14
Incompatibility with rust wayland-client crate


wez:
When trying to run [wezterm](https://github.com/wez/wezterm) under Wayland, EGL initialization fails with BAD_NATIVE_WINDOW.

Getting this under a debugger, it looks like this runtime check is the critical piece:

```
#1  wlEglIsWaylandWindowValid (window=window@entry=0x55555a44f740) at ../src/wayland-eglsurface.c:74
74          return WL_CHECK_INTERFACE_TYPE(surface, wl_surface_interface);
```

Poking around at the surface that was passed in:

```
>>> p surface
$5 = (struct wl_surface *) 0x555559938800
>>> p *(void**)surface
$8 = (void *) 0x555559904ea8 <wayland_client::protocol::wl_surface::wl_surface_interface>
>>> p *(struct wl_interface*)0x555559904ea8
$12 = {
  name = 0x555558d3994c "wl_surface",
  version = 4,
  method_count = 10,
  methods = 0x555559904d78 <wayland_client::protocol::wl_surface::wl_surface_requests>,
  event_count = 2,
  events = 0x555559904e78 <wayland_client::protocol::wl_surface::wl_surface_events>
}
```

And comparing with the `wl_surface` type that is checked against by this part of the macro:

https://github.com/NVIDIA/egl-wayland/blob/9c9c4c684983f7b0821f516f8821118170a785ea/include/wayland-eglutils.h#L60

```
>>> p &wl_surface_interface
$13 = (const struct wl_interface *) 0x7ffff7d4aa00 <wl_surface_interface>
>>> p wl_surface_interface
$14 = {
  name = 0x7ffff7d46a88 "wl_surface",
  version = 4,
  method_count = 10,
  methods = 0x7ffff7d4a540 <wl_surface_requests>,
  event_count = 2,
  events = 0x7ffff7d4a500 <wl_surface_events>
}
```

The interface appears to be compatible if the check were to perform a comparison of what's inside the `wl_interface`, even though the addresses are not identical.

Why aren't the addresses identical?

The rust client implementation employs code generation to produce most of its code; here's an example of the definition of that particular interface:

https://docs.rs/wayland-client/0.28.1/src/wayland_client/opt/rustwide/target/debug/build/wayland-client-fdd09ed62affdce6/out/wayland_api.rs.html#5885-5892

The rust client does, however, call out to the system libraries so that it should be comptible with EGL libraries.
[...]

PR #36, opened 2021-08-17
Fix interface type checks

danvd:
This does 3 things:

1. Makes WL_CHECK_INTERFACE_TYPE check interface pointers first because
   it's faster.
2. Adds 3rd parameter to WL_CHECK_INTERFACE_TYPE (string with interface
   name). Real interface names do not contain "_interface" postfix
in their names.
3. Removes dladdr() call from wlEglCheckInterfaceType and checks interface names directly.
   dladdr() returns DL_info, which in turn includes name of shared object that owns
   passed adress. And it's not interface name, but process or library path,
   e.g. "/usr/bin/alacritty" or "/usr/lib/libegl-renderer.so.1".

Fixes at least #34

### DIFF
--- a/include/wayland-eglutils.h
+++ b/include/wayland-eglutils.h
@@ -60,9 +60,9 @@ EGLBoolean wlEglFindExtension(const char *extension, const char *extensions);
 EGLBoolean wlEglPointerIsDereferencable(void *p);
 EGLBoolean wlEglCheckInterfaceType(struct wl_object *obj, const char *ifname);
 #ifndef WL_CHECK_INTERFACE_TYPE
-#define WL_CHECK_INTERFACE_TYPE(obj, ifname)                            \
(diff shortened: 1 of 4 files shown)


erik-kz:
Thanks for submitting the patch, and apologies for the slow review. While this would be sufficient to fix the rust issue, the problem is that because of the way external platform handles work, wlEglCheckInterfaceType needs to be safe to be called with an arbitrary pointer, even if it doesn't point to a wl_interface struct. Hence, it may not be safe to simply pass interface->name to strcmp since we can't be certain that memory is addressable.

To be safe, we need to check that we can indeed dereference a) interface->name, and b) the interface name itself. This is what I've done in https://github.com/NVIDIA/egl-wayland/commit/2eb4628d64a8297fb4c08aa796fdf33ff54a670f if you're curious.

danvd:
Got it. Thank for the fix!

Issue #52, opened 2022-04-04
Calling wl_egl_window_resize from different thread broken


sulix:
If wl_egl_window_resize() is called from a thread other than the one the window's EGLSurface is Current on, the state of which surface is current on which thread can get messed up.

See:
https://github.com/libsdl-org/SDL/pull/4821#issuecomment-943578513

Deferring the resize until eglSwapBuffers does appear to work as a workaround for this.

This breaks a number of games when run under Wayland, particularly:
- [Impostor Factory](https://store.steampowered.com/app/1182620/Impostor_Factory/)
- [Life is Strange](https://store.steampowered.com/app/319630/Life_is_Strange__Episode_1/?curator_clanid=36149206)

Both of which will hang on startup.

With SDL [using wayland by default](https://github.com/libsdl-org/SDL/commit/8ceba27d6291f1195e13608033ec439aec621fc6) now, this will start to affect any of these games which update to the latest SDL version.

sulix:
Implemented a hacky workaround which defers the resize until SwapBuffers here: https://github.com/NVIDIA/egl-wayland/pull/53

PR #53, opened 2022-04-04
Defer resizes until next SwapBuffers (work around #52)

sulix:
As explained in #52 and https://github.com/libsdl-org/SDL/pull/4821, several games are broken under nVidia/Wayland due to calling wl_egl_window_resize() from a different thread.

This change works around this by deferring the resize until eglSwapBuffers(), which fixes the aforementioned games.

I'm not sure if this is the right solution to this problem — and the implementation could use a lot of cleanup, even so, but it seems to be a marked improvement for me.

### DIFF
--- a/src/wayland-eglsurface.c
+++ b/src/wayland-eglsurface.c
@@ -1555,37 +1555,25 @@ EGLBoolean wlEglInitializeSurfaceExport(WlEglSurface *surface)
     return EGL_TRUE;
 }
 
-static void
-resize_callback(struct wl_egl_window *window, void *data)
+void
+wlEglResizeSurfaceIfRequired(WlEglDisplay *display, WlEglPlatformData *pData, WlEglSurface *surface)
 {
-    WlEglDisplay      *display = NULL;
-    WlEglPlatformData *pData   = NULL;
-    WlEglSurface      *surface = (WlEglSurface *)data;
     EGLint             err     = EGL_SUCCESS;
 
-    if (!window || !surface) {
+    if (!surface) {
         return;
     }
-
-    display = surface->wlEglDpy;
-    if (!wlEglIsWaylandDisplay(display->nativeDpy) ||
-        !wlEglIsWaylandWindowValid(surface->wlEglWin)) {
-        return;
-    }
-    pData = display->data;
 
     pthread_mutex_lock(&surface->mutexLock);
 
     /* Resize stream only if window geometry has changed */
-    if ((surface->width != window->width) ||
-        (surface->height != window->height) ||
-        (surface->dx != window->dx) ||
-        (surface->dy != window->dy)) {
+    if (surface->isResized) {
         // If a damage thread is in use, wait for it to finish processing all
         //   pending frames
         finish_wl_eglstream_damage_thread(surface, &surface->ctx, 0);
 
         discard_surface_context(surface);
+        surface->isResized = EGL_FALSE;
         surface->ctx.wlStreamResource = NULL;
         surface->ctx.isAttached = EGL_FALSE;
         surface->ctx.eglSurface = EGL_NO_SURFACE;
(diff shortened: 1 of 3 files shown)


src/wayland-eglsurface.c
```diff
@@ -1599,7 +1587,6 @@ resize_callback(struct wl_egl_window *window, void *data)
                                    surface,
                                    surface,
                                    pData->egl.getCurrentContext());
-
```
flibitijibibo, line 1602: Line can be added back to shrink the diff

src/wayland-eglswap.c
```diff
@@ -30,6 +30,8 @@
 #include <assert.h>
 #include <wayland-egl-backend.h>
 
+void wlEglResizeSurfaceIfRequired(WlEglDisplay *display, WlEglPlatformData *pData, WlEglSurface *surface);
```
flibitijibibo, line 33: Should this be added to `include/wayland-eglsurface.h`?

src/wayland-eglsurface.c
```diff
@@ -1611,6 +1598,37 @@ resize_callback(struct wl_egl_window *window, void *data)
     pthread_mutex_unlock(&surface->mutexLock);
 }
 
+
+
```
flibitijibibo, line 1602: Extra spaces

src/wayland-eglsurface.c
sulix: Done, thanks.

src/wayland-eglswap.c
sulix: Yup, I was going to do this, but forgot. Thanks!

sulix:
Pushed a new version with @flibitijibibo's suggestions and a couple of other minor style fixes. It's possible there's a nicer way of going about this still (maybe by saving off the desired size, and comparing that, rather than having an ``isResized`` variable, which I think is closer to what Mesa does…), but this seems to work for now.

As a totally unscientific possibility, this may have fixed (or made more rare) [this hang in KDE's plasmashell](https://bugs.kde.org/show_bug.cgi?id=448844). It's difficult to know for sure, as it happens quite randomly and rarely, but I haven't noticed it since making this change. Of course, plasmashell still has several other bugs, and this could be coincidence still…

erik-kz:
Thank you for the patch, and sorry for the very slow review. Apart from the one minor question above I think it looks good.

For what it's worth, I first noticed this issue in plasmashell which also does wl_egl_window_resize from a different thread and tried a very similar fix to this. However it didn't seem to be sufficient in that case. If I recall correctly, the reason was that resizing the surface right before eglSwapBuffers would discard any rendering done to the old surface, so you'd just get a black window unless / until something prompts it to redraw again.

However I imagine this isn't as much of an issue for SDL applications since they would typically be rendering every frame, so at the very least it's a strict improvement on the status quo.

When it was last discussed internally, I believe the consensus was that to really fix the issue properly we would need an EGL extension to allow you to explicitly resize a stream. There was a draft version of such a thing floating around but it never ended up getting finalized. But anyway, until that happens this workaround will be good to have.

sulix:
Thanks for having a look.

I'm afraid not sure what you mean by the "minor question above" — could you clarify?

As for the plasmashell issue, I'm pretty confident this patch is solving at least one of them: I was getting regular hangs before applying it, and am not anymore. However, it turns out there was also a [non-nVidia-specific Qt issue](https://codereview.qt-project.org/c/qt/qtwayland/+/393273), which had exactly the same symptoms of a hang. I've got that patch applied too, and no hangs. (Though, as mentioned, there are a few other cases of recently-resized things being mysteriously black and other visual corruption in Plasma, though nothing that seems like a regression…)

erik-kz:
> I'm afraid not sure what you mean by the "minor question above" — could you clarify?

Apologies, I was referring to my comment about the wl_surface_commit call, but I just realized it was still marked as "pending".

src/wayland-eglsurface.c
```diff
@@ -1611,6 +1599,36 @@ resize_callback(struct wl_egl_window *window, void *data)
...
+    pthread_mutex_lock(&surface->mutexLock);
+
+    /* Resize stream only if window geometry has changed */
+    if ((surface->width != window->width) ||
+        (surface->height != window->height) ||
+        (surface->dx != window->dx) ||
+        (surface->dy != window->dy)) {
+            surface->isResized = EGL_TRUE;
+            wl_surface_commit(surface->wlSurface);
```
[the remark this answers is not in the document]
sulix: I don't think so — there was something that needed it at some point in my initial experiments, but everything works without it now (and it doesn't look like the Mesa implementation commits here).

I've uploaded a new version without it.

sulix:
I've uploaded a new version with the ``wl_surface_commit`` call removed: everything still seems to work.

cubanismo (approved):
LGTM.

sulix:
Glad to hear this passed the tests.

I've been running this for a while, and haven't noticed anything either. I've only had plasmashell's rendering hang on me once since, and it was at the same time as a bunch of other applications died with messages like this, so I'm assuming its unrelated:
```
[drm:nv_drm_gem_alloc_nvkms_memory_ioctl [nvidia_drm]] *ERROR* [nvidia-drm] [GPU ID 0x00000100] Failed to allocate NVKMS memory for GEM object
steam[5144]: segfault at 0 ip 00000000f62f0b5d sp 00000000ff9a85c0 error 4 in libnvidia-glcore.so.510.60.02[f54c4000+1266000]
NVRM: Xid (PCI:0000:01:00): 31, pid=32703, Ch 00000030, intr 10000000. MMU Fault: ENGINE GRAPHICS GPCCLIENT_PROP_0 faulted @ 0x0_0489d000. Fault is of type FAULT_PDE ACCESS_TYPE_WRITE
```

PR #58, opened 2022-06-25
Remove mistakenly added wayland-eglswap.c.orig

sulix:
It looks like the ``src/wayland-eglswap.c.orig`` file was added to git by mistake in 53b6a87c27e8. Remove it again.


erik-kz:
My mistake, thank you!

Issue #57, opened 2022-06-23
Incomplete resizes with ≥1.1.10


foucault:
Greetings,
I've recently updated to egl-wayland 1.1.10 and I noticed that several applications do not resize properly in a Wayland session. Specifically part of the window is disappearing while actively resizing and it is only fully shown when the resize operation is finished. The examples below are for gnome-weather (GTK 4) but that also happens with QtQuick/Qt5 based applications (wayland backend) and also other applications like alacritty. Please note that in the second example the window is not actually black, this is probably an artifact of the screencap. The contents are still visible although not rearranged until after resizing finished. Latest build from git also exhibits the issue and occasionally the application segfaults outright. Perhaps unsurprisingly the first commit that exhibits the issue is [ddaa2720](https://github.com/NVIDIA/egl-wayland/commit/ddaa27209a8149c1960c70799c87151206401bbc). The issue is not specific to mutter as it can be replicated under KWin+Wayland.

My current system configuration
Distribution: Arch Linux
NVIDIA Driver version: 515.48.07
Kernel version: 5.18.5
GPU: NVIDIA 3070

[egl-wayland 1.1.9 - correct behaviour](https://user-images.githubusercontent.com/304223/175183505-b941807c-edb5-4739-a792-4a54658aba68.mp4)

[egl-wayland 1.1.10](https://user-images.githubusercontent.com/304223/175183597-c7503ad5-8c9c-41f2-aec2-2e6f7146c2f0.mp4)

ionenwks:
Hitting this as well while trying 515.57 + 1070 + gnome/mutter, and was rather bad on my end as it didn't really return to normal after finishing resize becoming unusable.

For now I'm using 1.1.10 with the aforementioned commit ddaa27209a8149c1960c70799c87151206401bbc reverted in Gentoo and everything seems(?) normal (not going back to 1.1.9 to keep wayland vulkan working with 515).

Bananaman:
Is this fixed now? Or maybe I'm not using the EGL Wayland backend.

My system has `egl-wayland-1.1.10-4.20220621git53b6a87.fc36` whose release notes specifically mention your bug report too:

https://bodhi.fedoraproject.org/updates/FEDORA-2022-e0d91b7c0b

I'm using Wayland with NVIDIA driver 510, and I have no issues while resizing windows.

It's possible that Fedora has removed the bad patch that you're talking about though...

foucault:
[...]

PR #60, opened 2022-07-09
Resize immediately if surface is current.

sulix:
Commit ddaa272 ("Defer resizes until next SwapBuffers") was perhaps overzealous, as most applications render and resize on the same thread. Given the resize operation destroys the current surface contents, deferring it to just before swapbuffers broke rendering for the first frame after a resize (see issue #57).

Instead, detect if the surface being resized is bound to the current thread, and only defer the resize if it is not. This is still not
technically correct (ideally we'd defer resizing the buffer until the next "safe point", when its contents are already being discarded, e.g. after a glClear()), but it preserves the old behaviour in the cases where it worked, but still fixes games which render from a different thread to the one which handles resizes. (I've tested this with Impostor Factory and Life is Strange, and both work, albeit still losing frame contents during resizes.)

This also should fix an issue whereby the wrong surface could be made current if the READ and DRAW surfaces differed, though I haven't found any affected real-world applications, there.

One slight curiosity is that Life Is Strange didn't enter fullscreen correctly (losing its window decorations, but only covering part of the screen) on KDE if I called ``eglGetCurrentSurface()`` each frame, before checking if ``surface`` was non-NULL (even though ``surface`` was never NULL). Given that it makes sense to only call ``eglGetCurrentSurface()`` when a resize is actually happening, it makes sense to have the code as-is anyway, but it still doesn't make sense. Maybe the optimiser's doing somthing weird (though I played around with locks and read/write barriers, and they didn't change anything…).

This was suggested by @erik-kz in https://github.com/NVIDIA/egl-wayland/issues/57#issuecomment-1179299387

### DIFF
--- a/src/wayland-eglsurface.c
+++ b/src/wayland-eglsurface.c
@@ -1616,7 +1616,9 @@ fail:
 void
 wlEglResizeSurfaceIfRequired(WlEglDisplay *display, WlEglPlatformData *pData, WlEglSurface *surface)
 {
-    EGLint             err     = EGL_SUCCESS;
+    EGLint             err             = EGL_SUCCESS;
+    WlEglSurface       *readSurface    = surface;
+    WlEglSurface       *drawSurface    = surface;
 
     if (!surface) {
         return;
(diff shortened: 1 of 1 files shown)


erik-kz:
Thanks for the fix, however I think we also need to address the fact that, when we do defer the resize, we will initially be presenting an undefined buffer. In https://github.com/NVIDIA/egl-wayland/commit/885f0a5e91fe9cfbfbcd98ff01f6b83503decef3 I've moved the resize to after the actual eglSwapBuffers call in such cases. Otherwise it ended up being basically the same as this patch (I also noticed the EGL_READ issue, so it will be good to have that fixed too).

sulix:
Thanks — I can confirm this fixes the issues I was seeing.

The one note I have is that moving the ``eglSwapBuffers`` call can break maximize on some windows under KDE/KWin. (The test case is MKXP games like Impostor Factory — run windowed, resize the window, maximize (or Alt+Enter to switch to fullscreen) — the window will be resized, but not moved.) There's a good chance that this is a KDE or SDL bug, though. It also only occurs with resizable windows which handle resizes in a different thread, which doesn't appear to be a common case, so I'm not particularly worried.

Cheers,
-- David

PR #71, opened 2022-10-21
[HACK] For Vulkan surfaces, assume opaque alpha composition.

flibitijibibo:
For more information, see the NVIDIA Developer Forums:

https://forums.developer.nvidia.com/t/wayland-vulkan-ignores-vkcompositealphaflagskhr/202227

This resolves the issue described in the above thread, but a potentially better fix would be for the driver to communicate this to egl-wayland explicitly.

### DIFF
--- a/src/wayland-eglsurface.c
+++ b/src/wayland-eglsurface.c
@@ -1021,8 +1021,14 @@ acquire_surface_image(WlEglDisplay *display, WlEglSurface *surface)
         /*
          * Before sending the format, check if we are ignoring alpha due to a
          * surface attribute.
+	 *
+	 * HACK: Also ignore alpha if this is a Vulkan surface.
+	 * Currently the best way to detect a Vulkan surface is 'isSurfaceProducer',
+	 * and the Vulkan driver only exposes OPAQUE_BIT for VK_KHR_surface.
+	 * If the driver exposes more bits, this check should be updated with it!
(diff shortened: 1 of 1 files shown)


erik-kz:
Thank you for the patch, however we've actually implemented support in our Vulkan driver for plumbing the "compositeAlpha" parameter from vkCreateSwapchainKHR though to egl-wayland (credit to Dogukan Korkmazturk). We've added a new version of wlEglCreateSurfaceExport which takes an attribute list to facilitate this, see https://github.com/NVIDIA/egl-wayland/commit/5d0194dcc7f4c334e0e71e8b5ec1b85311d72779

ALPHA_OPAQUE is the only value we actually advertise support for right now, so the end effect will be the same at this patch, but doing it that way does leave the door open for adding support for other values in the future.

The driver-side functionality will be available in the next major version 525.

flibitijibibo:
Saw the new beta driver release this morning! Just to confirm, does the 525 beta have this feature included? I didn't see it in the release notes so I just wanted to be sure...

erik-kz:
Sorry, I was mistaken in my earlier comment. It looks like the driver-side change just missed the cut-off for 525, so it will not be available until the following release, 530.

flibitijibibo:
Quick update to confirm that 530.41.03 has the fix for this issue.

PR #74, opened 2023-01-25
Destroy all created wayland object in terminateDisplay

Molytho:
These object are created but never destroyed. When the library is unload this leaves dangling pointers in libwayland-client leading to things like:
https://bugzilla.mozilla.org/show_bug.cgi?id=1759315

### DIFF
--- a/src/wayland-egldisplay.c
+++ b/src/wayland-egldisplay.c
@@ -628,6 +628,10 @@ static EGLBoolean terminateDisplay(WlEglDisplay *display, EGLBoolean globalTeard
             wl_eglstream_display_destroy(display->wlStreamDpy);
             display->wlStreamDpy = NULL;
         }
+        if (display->wlStreamCtl) {
+            wl_eglstream_controller_destroy(display->wlStreamCtl);
+            display->wlStreamCtl = NULL;
+        }
         if (display->wpPresentation) {
             wp_presentation_destroy(display->wpPresentation);
             display->wpPresentation = NULL;


erik-kz:
Thank you very much for catching this bug, and for sharing a fix. Austin's recent change 7af7082eac2170da66990edda2ca73a221e2086e should already resolve the issue with zwp_linux_dmabuf_v1, but we definitely should also be destroying the wl_eglstream_controller interface. Could you please update the patch to remove the now-duplicated part...
```
       if (display->wlDmaBuf) {
            zwp_linux_dmabuf_v1_destroy(display->wlDmaBuf);
            display->wlDmaBuf = NULL;
        }
```

erik-kz:
Thanks, just need to get a +1 from another NVIDIA engineer and then I will merge.

kki0t:
We would appreciate a new release of egl-wayland, because this bug, is still present in latest 1.1.11.

This bug causes some pretty nice chaos, especially in Firefox and its `glxtest` component on Wayland.
`glxtest` SIGSEGVs because of it, which Firefox notices, and thus interprets as no GPU found, if I understand it.
It then falls back to software rendering.
See [Bugzilla #1759315](https://bugzilla.mozilla.org/show_bug.cgi?id=1759315)
Note that it seemingly doesn't happen everywhere. IIRC this is a non-issue on e.g. Mutter.

I think this is the place to ask. Sorry if I'm wrong. 
Meanwhile I'll try to build from git master if I can make that _one third-party tool_ work with it.

TLATER:
*Really* would appreciate a release for this. Firefox is completely unable to run webgl because of this, so lots of websites are flat out broken.

I was hoping a release would have been around by now, given your previous release schedule seemed to include May occasionally - what cadence should we expect?

erik-kz:
Why does a warning break WebGL? That seems like an problem with Firefox..

Anyway, I'm planning to cut a new release to coincide with the 535 driver release at the end of the month.

PR #79, opened 2023-04-07
Fix wrong destruction order for event queue

q234rty:
libwayland 1.22 added a [warning](https://gitlab.freedesktop.org/wayland/wayland/-/blob/1.22.0/src/wayland-client.c?ref_type=tags#L308) that could currently be triggered by running `eglgears_wayland` under NVIDIA wayland:
```
warning: queue 0x5560c4ce1b30 destroyed while proxies still attached:
  wl_registry@3 still attached
```
Firefox currently [crashes](https://bugzilla.mozilla.org/show_bug.cgi?id=1826583) upon such warnings. Mesa has fixed a similar bug in https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/21646

### DIFF
--- a/src/wayland-egldisplay.c
+++ b/src/wayland-egldisplay.c
@@ -704,11 +704,11 @@ static void getServerProtocolsInfo(struct wl_display *nativeDpy,
         }
     }
 
+    if (wlRegistry) {
+        wl_registry_destroy(wlRegistry);
+    }
     if (queue) {
         wl_event_queue_destroy(queue);
-    }
-    if (wlRegistry) {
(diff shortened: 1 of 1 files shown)


AlynxZhou:
I tested this with a NVIDIA 2070 Super using nvidia driver, it fixes the warning.

colemickens:
Ping, if this can be merged/tagged it would be great. It's currently potentially blocking distro(s) that want to update libwayland.

erik-kz:
Thanks. This looks good to me. Will merge after getting +1 from another NVIDIA engineer.

PR #100, opened 2024-01-14
egl-wayland: Read compositor device from zwp_linux_dmabuf_v1

Molytho:
Compositors start dropping support for wl_drm because it is replaced by linux-dmabuf. See https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/2f2a55ebd3c5a950f1bf8c280edf0838721363da. Try reading compositors main device from linux-dmabuf v4 if supported and otherwise fallback to wl_drm.

### DIFF
--- a/src/wayland-egldisplay.c
+++ b/src/wayland-egldisplay.c
@@ -1186,7 +1186,7 @@ fail:
     free(eglDeviceList);
     free(protocols.drm_name);
 
-    if (display->ownNativeDpy) {
+    if (display && display->ownNativeDpy) {
         wl_display_disconnect(display->nativeDpy);
     }
     free(display);
--- a/src/wayland-egldisplay.c
+++ b/src/wayland-egldisplay.c
@@ -39,17 +39,23 @@
 #include <fcntl.h>
 #include <sys/mman.h>
 #include <xf86drm.h>
+#include <dlfcn.h>
 
 typedef struct WlServerProtocolsRec {
     EGLBoolean hasEglStream;
     EGLBoolean hasDmaBuf;
-    EGLBoolean hasDrm;
-    struct wl_drm *wldrm;
+    struct zwp_linux_dmabuf_v1 *wlDmaBuf;
+    dev_t devId;
+
+    struct wl_drm *wlDrm;
     char *drm_name;
 } WlServerProtocols;
 
 /* TODO: Make global display lists hang off platform data */
 static struct wl_list wlEglDisplayList = WL_LIST_INITIALIZER(&wlEglDisplayList);
+
+static bool getDeviceFromDevIdInitialised = false;
+static int (*getDeviceFromDevId)(dev_t dev_id, uint32_t flags, drmDevice **device) = NULL;
 
 EGLBoolean wlEglIsWaylandDisplay(void *nativeDpy)
 {
@@ -501,6 +507,96 @@ static const struct wl_drm_listener drmListener = {
 
 
 static void
+dmabuf_feedback_check_main_device(void *data,
+                            struct zwp_linux_dmabuf_feedback_v1 *dmabuf_feedback,
+                            struct wl_array *dev)
+{
+    WlServerProtocols *protocols = (WlServerProtocols *)data;
+    (void) dmabuf_feedback;
+
+    assert(dev->size == sizeof(dev_t));
+    memcpy(&protocols->devId, dev->data, sizeof(dev_t));
+}
+
+static void
+dmabuf_feedback_check_tranche_target_device(void *data,
+                                      struct zwp_linux_dmabuf_feedback_v1 *dmabuf_feedback,
+                                      struct wl_array *dev)
+{
+    (void) data;
+    (void) dmabuf_feedback;
+    (void) dev;
+}
+
+static void
+dmabuf_feedback_check_tranche_flags(void *data,
+                              struct zwp_linux_dmabuf_feedback_v1 *dmabuf_feedback,
+                              uint32_t flags)
+{
+    (void) data;
+    (void) dmabuf_feedback;
+    (void) flags;
+}
+
+static void
+dmabuf_feedback_check_tranche_formats(void *data,
+                                struct zwp_linux_dmabuf_feedback_v1 *dmabuf_feedback,
+                                struct wl_array *indices)
+{
+    (void) data;
+    (void) dmabuf_feedback;
+    (void) indices;
+}
+
+static void
+dmabuf_feedback_check_tranche_done(void *data,
+                             struct zwp_linux_dmabuf_feedback_v1 *dmabuf_feedback)
+{
+    (void) data;
+    (void) dmabuf_feedback;
+}
+
+static void
+dmabuf_feedback_check_done(void *data, struct zwp_linux_dmabuf_feedback_v1 *dmabuf_feedback)
+{
+    WlServerProtocols *protocols = (WlServerProtocols *)data;
+    (void) dmabuf_feedback;
+
+    assert(protocols->devId != 0);
+    drmDevice *drm_device;
+    assert(getDeviceFromDevId);
+    if (getDeviceFromDevId(protocols->devId, 0, &drm_device) == 0) {
+        if (drm_device->available_nodes & (1 << DRM_NODE_RENDER)) {
+            protocols->drm_name = strdup(drm_device->nodes[DRM_NODE_RENDER]);
+        }
+
+        drmFreeDevice(&drm_device);
+    }
+}
+
+static void
+dmabuf_feedback_check_format_table(void *data,
+                             struct zwp_linux_dmabuf_feedback_v1 *dmabuf_feedback,
+                             int32_t fd, uint32_t size)
+{
+    (void) data;
+    (void) dmabuf_feedback;
+    (void) fd;
+    (void) size;
+}
+
+static const struct zwp_linux_dmabuf_feedback_v1_listener dmabuf_feedback_check_listener = {
+        .done = dmabuf_feedback_check_done,
+        .format_table = dmabuf_feedback_check_format_table,
+        .main_device = dmabuf_feedback_check_main_device,
+        .tranche_done = dmabuf_feedback_check_tranche_done,
+        .tranche_target_device = dmabuf_feedback_check_tranche_target_device,
+        .tranche_formats = dmabuf_feedback_check_tranche_formats,
+        .tranche_flags = dmabuf_feedback_check_tranche_flags,
+};
+
+
+static void
 registry_handle_global_check_protocols(
                        void *data,
                        struct wl_registry *registry,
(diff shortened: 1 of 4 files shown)


erik-kz:
Thanks for the patch! Note, however, that we already bind the `zwp_linux_dmabuf_v1` protocol and retrieve the default feedback during `wlEglInitializeHook`. So we shouldn't need to do it a second time.

I *do* think it's a good idea to use `zwp_linux_dmabuf_v1` to determine the compositor device instead of `wl_drm` when possible. If you could modify the patch to keep that part but use the existing `WlEglDisplay::defaultFeedback` I would be happy to merge it.

Molytho:
I changed the code to bind to zwp_linux_dmabuf_v1 only once.
I also added some error handling to getServerProtocolsInfo.
Sadly this made it more ugly

src/wayland-egldisplay.c
```diff
@@ -435,8 +436,9 @@ registry_handle_global(void *data,
         /*
          * Version 3 added format modifier support, which the dmabuf
          * support in this library relies on.
+         * If the version is >= 4 we already bound it in getServerProtocolsInfo
```
erik-kz, line 439: I think it might simplify thing a little bit to always bind `zwp_linux_dmabuf_v1` during `eglGetPlatformDisplay`, regardless of whether version 3 or 4 is supported. Of course, if version 4 isn't supported we also need to bind `wl_drm`. Then we can get rid of the code that binds it during `eglInitialize`.

src/wayland-egldisplay.c
```diff
@@ -670,48 +674,103 @@ EGLBoolean wlEglTerminateHook(EGLDisplay dpy)
...
     wl_proxy_set_queue((struct wl_proxy *)wrapper, queue);
 
     /* Listen to wl_registry events and make a roundtrip in order to find the
      * wl_eglstream_display global object.
      */
     wlRegistry = wl_display_get_registry(wrapper);
+    if (!wlRegistry) {
+        goto out;
+    }
     wl_proxy_wrapper_destroy(wrapper); /* Done with wrapper */
```
erik-kz, line 722: I think we can remove this (and the line below) since we're now destroying the wrapper at the end.

src/wayland-egldisplay.c
```diff
@@ -898,7 +960,7 @@ EGLDisplay wlEglGetPlatformDisplayExport(void *data,
         }
     }
 
-    if (!protocols.hasDrm || (!protocols.hasEglStream && !protocols.hasDmaBuf)) {
+    if (!protocols.hasEglStream && !protocols.hasDmaBuf) {
```
erik-kz, line 963: Do we really need to change this condition? If we don't have `zwp_linux_dmabuf_v1`  or `wl_eglstream` but we do have `wl_drm` I think we're still fine, are we not?

src/wayland-egldisplay.c
```diff
@@ -670,48 +674,103 @@ EGLBoolean wlEglTerminateHook(EGLDisplay dpy)
...
 
-static void getServerProtocolsInfo(struct wl_display *nativeDpy,
+static char* getNodeFromDrmDevice(dev_t dev) {
+    char *result = NULL;
+    drmDevice *drm_device;
+    if (drmGetDeviceFromDevId(dev, 0, &drm_device) == 0) {
+        if (drm_device->available_nodes & (1 << DRM_NODE_RENDER)) {
+            result = strdup(drm_device->nodes[DRM_NODE_RENDER]);
+        } else if (drm_device->available_nodes & (1 << DRM_NODE_PRIMARY)) {
+            result = strdup(drm_device->nodes[DRM_NODE_PRIMARY]);
```
erik-kz, line 684: In `wlEglGetPlatformDisplayExport` we only check if the render node of an `EGLDevice` matches `drm_name`. Therefore I don't think there's any point returning a primary node path from this function. If there's no render node available we can just return NULL.

src/wayland-egldisplay.c
Molytho: We definitely have to change this because this always results in a failure if wl_drm is not available.
I don't think that we are fine with wl_drm alone. This if statement currently leads to a fail if neither zwp_linux_dmabuf_v1 nor wl_eglstream are available

src/wayland-egldisplay.c
Molytho: Sounds good. This sadly still leaves on issue:
if the display is initialized and then destroyed we also free all wayland objects in terminateDisplay. The spec allows initializing the same display again so we would still need to bind the interfaces in eglInitialize if they aren't bound already.

src/wayland-egldisplay.c
erik-kz: Oh, yeah you're right. Sorry, I should have thought it through more carefully.

src/wayland-egldisplay.c
erik-kz: Ah shit, true, I had forgotten about that. Well, if we might need to re-bind the interface during `eglInitialize` anyway maybe it would be better to just go with the original version of your patch. Now I feel like a bit of an ass for making you re-write it, really sorry about that.

Molytho: No problem. I initially also forgot about that case.
I reverted to the initial patch but kept the new error handling in getServerProtocolsInfo

erik-kz:
Thanks. This looks good to me. I just need to get a +1 from another NVIDIA engineer and then I will merge it.

src/wayland-egldisplay.c
```diff
@@ -500,6 +502,95 @@ static const struct wl_drm_listener drmListener = {
...
+
+static void
+dmabuf_feedback_check_done(void *data, struct zwp_linux_dmabuf_feedback_v1 *dmabuf_feedback)
+{
+    WlServerProtocols *protocols = (WlServerProtocols *)data;
+    (void) dmabuf_feedback;
+
+    assert(protocols->devId != 0);
+    drmDevice *drm_device;
+    if (drmGetDeviceFromDevId(protocols->devId, 0, &drm_device) == 0) {
```
cubanismo, line 563: This is a relatively new function in libdrm. I don't see a particular version of libdrm required in the  current meson.build. Perhaps we should add a requirement for version 2.4.109, the first version to include this function?

src/wayland-egldisplay.c
```diff
@@ -670,48 +763,74 @@ EGLBoolean wlEglTerminateHook(EGLDisplay dpy)
...
+                wl_display_roundtrip_queue(nativeDpy, queue);
+                zwp_linux_dmabuf_feedback_v1_destroy(default_feedback);
+            }
+        } else if (protocols->wlDrm) {
+            wl_drm_add_listener(protocols->wlDrm, &drmListener, protocols);
             wl_display_roundtrip_queue(nativeDpy, queue);
-            /* destroy our wl_drm object */
-            wl_drm_destroy(protocols->wldrm);
+        }
+        result = protocols->drm_name != NULL;
```
cubanismo, line 813: I see some discussion of this, but is this really fatal? I thought the streams path worked fine without this.

cubanismo (commented):
A suggestion and a question inline.

src/wayland-egldisplay.c
erik-kz: I think it kinda doesn't matter since `checkNvidiaDrmDevice` will return false if `protocols->drm_name` is NULL, meaning we will fail `eglGetPlatformDisplay` anyway unless `__NV_PRIME_RENDER_OFFLOAD` is set. But PRIME render-offload definitely doesn't doesn't work with the EGLStreams path.

src/wayland-egldisplay.c
kbrenneman: Are there any systems that are both new enough to support Wayland and old enough to not have `drmGetDeviceFromDevId`?

If so, then we'd probably need to look up `drmGetDeviceFromDevId` at runtime using dlsym.

erik-kz: 2.4.109 was released May of 2021, so maybe? But if we're going to do that would it be easier to just return `"/dev/dri/renderD" + minor(protocols->devId)`? It looks like that's pretty much how libdrm populates `drmDevice::nodes`.

src/wayland-egldisplay.c
cubanismo: Hrm. Did we regress that as part of the not-waking-up-GPUs-for-every-eglGetDisplay() change? It seems like if so, we sort of broke the eglstreams path. IIRC the wl_drm usage was added for the dma-buf path, so I doubt the eglstreams path was intended to require it.

src/wayland-egldisplay.c
cubanismo: Even though I think that's reliable in practice, I don't think we're supposed to rely on file layout assumptions. It probably is safe to assume systems without DRM 2.4.109 are running compositors that still support wl_drm though, so a dynamic lookup and skipping this path if it isn't there seems fine. Normally I'd say just compile it out if the DRM version is too old, but since we'll have to ship this as a binary in our driver package as well, I agree a runtime check is needed.

src/wayland-egldisplay.c
erik-kz: The EGLStreams path only works if the compositor calls `eglBindWaylandDisplayWL`, since that's where we bind the `wl_eglstream` interface. But in that case we will also bind our own implementation of `wl_drm`. So we shouldn't really need to worry about the EGLStreams path working without `wl_drm`, it will always be available.

To your question, though, no I don't think the change you mentioned regressed anything. All it did was move the failure point a little bit earlier. But even before that change, if we couldn't determine what device the compositor was using we would  fail (unless `__NV_PRIME_RENDER_OFFLOAD` is set).

cubanismo: OK. perhaps I'm misremembering.

src/wayland-egldisplay.c
Molytho: I updated the code to use dlsym now

src/wayland-egldisplay.c
erik-kz: I mean, there *was* a time when we would just choose a device completely arbitrarily, with no concern for which device the compositor was using. But we needed to change that because it was problematic on PRIME systems. See https://github.com/NVIDIA/egl-wayland/commit/d4937adc5cd04ac7df98fc5616e40319fb52fdee

src/wayland-egldisplay.c
```diff
@@ -670,48 +768,79 @@ EGLBoolean wlEglTerminateHook(EGLDisplay dpy)
...
+        goto done;
+    }
     ret = wl_registry_add_listener(wlRegistry,
                                    &registryListener,
                                    protocols);
     if (ret == 0) {
         wl_display_roundtrip_queue(nativeDpy, queue);
-        if (protocols->hasDrm) {
+        if (!getDeviceFromDevIdInitialised) {
+            getDeviceFromDevId = dlsym(RTLD_DEFAULT, "drmGetDeviceFromDevId");
```
cubanismo, line 807: Thanks for adding this, but I think you also need to update the meson and automake/autoconf files since this is the first use of libdl in egl-wayland AFAIK. For an example of the meson changes, see the egl-gbm equivalent:

https://github.com/NVIDIA/egl-gbm/blob/b7d9f41a959c00d83d948ec1a17a287f3ce7f9aa/src/meson.build#L1

I'm not familiar enough with autoconf myself to make a suggestion there.

erik-kz: We actually used to link against libdl. You could just revert the changes to Makefile.am, configure.ac, and Meson.build made by  https://github.com/NVIDIA/egl-wayland/commit/2eb4628d64a8297fb4c08aa796fdf33ff54a670f

Molytho: Thanks. This was quiet helpful. Should be done now.

cubanismo: Awesome, thanks.

cubanismo (approved):
Latest version LGTM.

PR #107, opened 2024-04-12
Fix possible build error with RTLD_DEFAULT

kbrenneman:
You have to define `_GNU_SOURCE` to use `RTLD_DEFAULT` (at least with glibc), so this adds that macro to the Meson build.

As far as I can tell, the autotools scripts already set _GNU_SOURCE by default, so that doesn't need to be changed.

I also added an #ifdef around where `RTLD_DEFAULT` is used, so it'll be able to build on systems where RTLD_DEFAULT isn't provided at all.


src/meson.build
```diff
@@ -8,6 +8,7 @@ add_project_arguments('-Wall', language : 'c')
 add_project_arguments('-Werror', language : 'c')
 add_project_arguments('-fvisibility=hidden', language : 'c')
 add_project_arguments('-DWL_HIDE_DEPRECATED', language : 'c')
+add_project_arguments('-D_GNU_SOURCE', language : 'c')
```
cubanismo, line 11: It's not clear to me autotools always enables _GNU_SOURCE. It might be getting enabled as a side effect of something else, but not reliably. A little Googling found this, for example:

https://reversed.top/2013-10-16/autoconf-how-to-define-gnu-source-correctly/

src/wayland-egldisplay.c
```diff
@@ -815,7 +815,9 @@ static bool getServerProtocolsInfo(struct wl_display *nativeDpy,
     if (ret == 0) {
         wl_display_roundtrip_queue(nativeDpy, queue);
         if (!getDeviceFromDevIdInitialised) {
+#if defined(RTLD_DEFAULT)
```
cubanismo, line 818: This doesn't seem desirable. I think it's better to fail the build if this isn't defined. Then we can catch issues like the one you're fixing at build time, rather than wonder what's going on at runtime when something seems a little off.

cubanismo (changes requested):

src/meson.build
kbrenneman: Ah, yeah, that's a good point. I had based that on some local testing, but if there's an autoconf macro for that, then we should use it.

src/wayland-egldisplay.c
kbrenneman: I had figured that the #ifdef would let it run on systems that don't have RTLD_DEFAULT, but now that I think about it more, if RTLD_DEFAULT is missing, it's more likely that there's a problem with the meson/configure scripts. In which case, yes, a loud build-time failure would be preferable.

cubanismo (approved):

PR #109, opened 2024-05-15
egl-wayland: Validate and use the passed value of EGL_EXT_present_opaque

Kontrabant:
It was assumed that the presence of the EGL_PRESENT_OPAQUE_EXT enum in the surface attribute list meant that the alpha channel should be ignored as if it was always followed by a value of EGL_TRUE, however, according to the spec, valid usage allows for a value of EGL_FALSE as well, which should result in the alpha channel being used for blending.

Validate and use the value instead of assuming that the presence of the extension enum value implies that it is enabled.


erik-kz:
Thanks for the patch, it looks good to me. Just need a +1 from another NVIDIA engineer, @amshafer would you be able to take a look?

amshafer:
Looks good to me as well, thanks

PR #113, opened 2024-07-01
egl-wayland: Fix memory leak in wlEglCreateSurfaceExport

jexposit:
The "surface" variable is not freed in these 2 cases.

Fixes: 63e82b06fbc0 ("wayland-egl: avoid unnecessary commit")

---

Replacing that code branch with a `goto fail` might work as well. The maintainers will know better :)


dkorkmazturk:
Thanks for this MR! However, this function has been deprecated since 545 and will be removed in an upcoming release.

amshafer:
Closing since this will be removed in the near future.

PR #120, opened 2024-07-26
egl-wayland: Fix an unbounded array growth issue

dkorkmazturk:
We do not seem to shrink the dynamically allocated streamImages array that we use for storing the resources associated with the swapchain images. When an entry in this array is destroyed in the destroy_stream_image() function, its fields are simply reverted to default/invalid values. Later on, the entries in this array are tried to be recycled in the add_surface_image() function. However, this approach is error-prone since we keep valid and invalid entries together, which requires us to check the validity of the entries every time we access them. Also, when explicit sync is in use, this array just keeps growing over time, especially during the application window resizes, due to a bug in the entry destruction logic.

So, to solve these problems, this change converts the dynamically allocated streamImages array into a linked list for simplified insertions and deletions. Each entry in this linked list is removed from the list and deallocated once they are no longer needed. So, all of the entries in the list stay valid.

Per-entry mutexes are replaced with a single mutex that guards accesses to the entire list to make sure that the linked list does not get corrupted when it is accessed from multiple threads. This only happens when explicit sync is not in use. The sizes of the critical sections that are protected by this new mutex are very small. To test to see if this change creates lock contention issues, weston-simple-egl application with swap interval of 0 was run on a Wayland compositor that does not support explicit sync to create as much lock contention as possible. However, no measurable difference in performance was observed after this change was applied.

As a side effect of this change, a bug in the
wlEglSurfaceCheckReleasePoints() function, where we wrongly assumed that all the entries in the streamImages array were valid, is fixed. This bug caused us to pass destroyed DRM syncobjs to the drmSyncobjTimelineWait() function, which led to random application freezes as a result in some cases since it prevented images from being released back to the EGL stream.

Another side effect of this change is that, it makes the maximum number of entries in this list known when explicit sync is in use, allowing us to avoid dynamically allocating the arrays for the list of DRM syncobjs and timeline points in wlEglSurfaceCheckReleasePoints(). This fixes a memory leak issue that can happen if only one of these allocations fails.

### DIFF
--- a/src/wayland-eglsurface.c
+++ b/src/wayland-eglsurface.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014-2022, NVIDIA CORPORATION. All rights reserved.
+ * Copyright (c) 2014-2024, NVIDIA CORPORATION. All rights reserved.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a
  * copy of this software and associated documentation files (the "Software"),
@@ -47,6 +47,7 @@
 #include <stdio.h>
 
 #define WL_EGL_WINDOW_DESTROY_CALLBACK_SINCE 3
+#define MAX_IMAGES 4 /* The swapchain image count */
 
 enum BufferReleaseThreadEvents {
     BUFFER_RELEASE_THREAD_EVENT_TERMINATE,
@@ -565,25 +566,25 @@ destroy_stream_image(WlEglDisplay *display,
     WlEglPlatformData   *data     = display->data;
     EGLDisplay           dpy      = display->devDpy->eglDisplay;
 
-    /* Must be called with image->mutex already locked */
-
-    assert(image->eglImage != EGL_NO_IMAGE_KHR);
-    data->egl.destroyImage(dpy, image->eglImage);
-    image->eglImage = EGL_NO_IMAGE_KHR;
-    image->releasePoint = 0;
+    /* Must be called with surface->ctx.streamImagesMutex already locked */
 
     if (surface->ctx.currentBuffer == image->buffer) {
         surface->ctx.currentBuffer = NULL;
     }
 
-    if (image->buffer && !image->attached) {
-        wl_buffer_destroy(image->buffer);
-        image->buffer = NULL;
+    if (!surface->wlSyncobjSurf && image->attached) {
+        // This is used for delaying the destruction of images only when
+        // explicit-sync is not in use to prevent the buffer release thread
+        // from accessing images after they are deallocated.
+        image->destructionPending = EGL_TRUE;
+        return;
     }
 
-    if (!wl_list_empty(&image->acquiredLink)) {
-        wl_list_remove(&image->acquiredLink);
-        wl_list_init(&image->acquiredLink);
+    assert(image->eglImage != EGL_NO_IMAGE_KHR);
+    data->egl.destroyImage(dpy, image->eglImage);
+
+    if (image->buffer) {
+        wl_buffer_destroy(image->buffer);
     }
 
     if (image->wlReleaseTimeline) {
@@ -591,10 +592,13 @@ destroy_stream_image(WlEglDisplay *display,
         drmSyncobjDestroy(display->drmFd, image->drmSyncobjHandle);
         if (image->acquireSync != EGL_NO_SYNC_KHR) {
             data->egl.destroySync(dpy, image->acquireSync);
-            image->acquireSync = EGL_NO_SYNC_KHR;
         }
-        image->releasePending = false;
     }
+
+    wl_list_remove(&image->acquiredLink);
+    wl_list_remove(&image->link);
+
+    free(image);
 }
 
 static void
@@ -606,23 +610,12 @@ destroy_surface_context(WlEglSurface *surface, WlEglSurfaceCtx *ctx)
     EGLSurface         surf     = ctx->eglSurface;
     EGLStreamKHR       stream   = ctx->eglStream;
     void              *resource = ctx->wlStreamResource;
-    uint32_t           i;
 
     finish_wl_eglstream_damage_thread(surface, ctx, 1);
 
     ctx->eglSurface       = EGL_NO_SURFACE;
     ctx->eglStream        = EGL_NO_STREAM_KHR;
     ctx->wlStreamResource = NULL;
-
-    for (i = 0; i < ctx->numStreamImages; i++) {
-        WlEglStreamImage *image = surface->ctx.streamImages[i];
-
-        pthread_mutex_lock(&image->mutex);
-        if (image->eglImage != EGL_NO_IMAGE_KHR) {
-            destroy_stream_image(display, surface, image);
-        }
-        pthread_mutex_unlock(&image->mutex);
-    }
 
     if (surf != EGL_NO_SURFACE) {
         data->egl.destroySurface(dpy, surf);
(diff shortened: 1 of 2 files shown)


src/wayland-eglsurface.c
```diff
@@ -1199,14 +1201,20 @@ wlEglSurfaceCheckReleasePoints(WlEglDisplay *display, WlEglSurface *surface)
      * Note that there are some bugs with older kernels where this may not
      * signal correctly.
      */
-    if (drmSyncobjTimelineWait(display->drmFd, syncobjs, syncPoints,
-                               surface->ctx.numStreamImages, timeout,
-                               DRM_SYNCOBJ_WAIT_FLAGS_WAIT_AVAILABLE,
-                               &firstSignaled) != 0) {
-        goto end;
+    {
```
amshafer, line 1204: What's the purpose of this bracketed block now that the if condition is gone? Is it just so `syncobjWaitRes` has a bounded lifetime? If so I think we should get rid of the extra brackets and declare `syncobjWaitRes` above with the rest of the local variables.

src/wayland-eglsurface.c
```diff
@@ -1291,19 +1293,25 @@ acquire_surface_image(WlEglDisplay *display, WlEglSurface *surface)
...
+                found = EGL_TRUE;
+                break;
+            }
         }
-    }
 
-    if (!image) {
-        goto fail_destroy_sync;
+        if (!found) {
+            pthread_mutex_unlock(&surface->ctx.streamImagesMutex);
```
amshafer, line 1310: I think it would make sense to move this `pthread_mutex_unlock` into `fail_destroy_sync`. This matches how we unlock the mutex in `fail_release` and also prevents us from accidentally forgetting to unlock it if we use `fail_destroy_sync` from a second location in the future.

src/wayland-eglsurface.c
```diff
@@ -2474,25 +2440,19 @@ EGLBoolean wlEglSurfaceRef(WlEglDisplay *display, WlEglSurface *surface)
...
         return;
     }
 
     wlEglMutexDestroy(&surface->mutexLock);
+    wlEglMutexDestroy(&surface->ctx.streamImagesMutex);
 
     if (!surface->ctx.isOffscreen) {
         wlEglMutexDestroy(&surface->mutexFrameSync);
         pthread_cond_destroy(&surface->condFrameSync);
     }
```
amshafer, line 2455: We no longer have to clean up the stream images array, but maybe it would be a good idea to assert that the `wl_list` is empty? By inspection it looks like we always clean up the surface context properly but an assert here could be helpful catching any future issues.

src/wayland-eglsurface.c
```diff
@@ -2564,26 +2523,41 @@ static EGLBoolean wlEglDestroySurface(EGLDisplay dpy, EGLSurface eglSurface)
...
+        finish_wl_buffer_release_thread(surface);
 
-    finish_wl_buffer_release_thread(surface);
+        /*
+         * If there are remaining images in the streamImages list, destroy them
+         * here safely since the buffer release thread is no longer running.
+         */
+        wl_list_for_each_safe(image, nextImage, &surface->ctx.streamImages, link) {
+            destroy_stream_image(display, surface, image);
+        }
```
amshafer, line 2558: I'm a little confused about the flow here. Shouldn't we only do the attached buffer cancelling and `finish_wl_buffer_release_thread` above when the buffer release thread is in use, but always destroy all the stream images here? Won't this leak them on explicit sync? Or do we know for sure that they will all have been destroyed by this point on explicit sync?

If we know they should all be destroyed it seems like we should mention that in the comments about why this isn't needed on explicit sync?

src/wayland-eglsurface.c
```diff
@@ -1135,40 +1144,33 @@ wlEglSurfaceCheckReleasePoints(WlEglDisplay *display, WlEglSurface *surface)
...
             numSyncPoints++;
-        } else {
-            /*
-             * Use a bogus point for acquired images so we can keep our indices
-             * the same. This won't affect anything since it will never have a fence
-             * appear.
-             */
-            syncPoints[i] = UINT64_MAX;
+
+            if (numSyncPoints >= MAX_IMAGES) {
```
kbrenneman, line 1169: This is incorrect if we have exactly MAX_IMAGES elements, because we'll have already incremented numSyncPoints to MAX_IMAGES.

Instead, this check needs to go before the assignment and increment above.

Or alternatively, we could have an extra loop to count the number of elements we'll need and then call `alloca` to allocate the arrays for them.

dkorkmazturk: numSyncPoints is incremented to n right after the elements at the index n-1 is assigned. We check for this condition right after the increment and before the loop which we assign the elements at the index n. So as long as the initial value of numSyncPoints is less than MAX_IMAGES, which currently is, this should work. Though I believe moving the check above the assignment as you suggested would be less error-prone. We wouldn't be relying on an assumption of the initial value of numSyncPoints.

src/wayland-eglsurface.c
dkorkmazturk: I believe it is a good practice to keep the scope of the variables minimal. The rest of the code in this function does not need to know the returned value.  But I do not mind removing this scope and moving the declaration of the variable if you prefer that way.

src/wayland-eglsurface.c
```diff
@@ -1291,19 +1293,25 @@ acquire_surface_image(WlEglDisplay *display, WlEglSurface *surface)
...
     }
 
-    for (i = 0; i < surface->ctx.numStreamImages; i++) {
-        if (surface->ctx.streamImages[i]->eglImage == eglImage) {
-            image = surface->ctx.streamImages[i];
-            break;
+    pthread_mutex_lock(&surface->ctx.streamImagesMutex);
+
+    // Locate the corresponding WlEglStreamImage
+    {
```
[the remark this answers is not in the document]
dkorkmazturk: Similarly, scope reduction. The rest of the code does not need to know if the image is found or not.

src/wayland-eglsurface.c
dkorkmazturk: An assert is already added to the line 2560.

src/wayland-eglsurface.c
dkorkmazturk: destroy_surface_context() call at the beginning of this function should take care of the stream image destruction when explicit sync is in use. I will add a comment mentioning it.

src/wayland-eglsurface.c
kbrenneman: Limiting variable scope is nice when you can do it, but not at the readability cost of disrupting the block structure.

amshafer: Stylistically this isn't something we really seem to do, but I think if you want to attain the same scope reduction we could create a small helper function to wrap `drmSyncobjTimelineWait`. I think that could look nicer than this block scoping

kbrenneman: In this case, you could also remove `syncobjWaitRes` entirely and just check `errno` in your assert.

dkorkmazturk: Checking errno makes sense instead of using a separate variable, which helps to clean this up. Thank you for all the suggestions.

src/wayland-eglsurface.c
dkorkmazturk: I believe jumping to fail_release actually makes more sense since the eglImage here is already acquired. It also looks like fail_release needed some changes for how image is handled now.

src/wayland-eglsurface.c
```diff
@@ -2564,26 +2517,45 @@ static EGLBoolean wlEglDestroySurface(EGLDisplay dpy, EGLSurface eglSurface)
...
         /*
          * Destroy any attached buffers to ensure no further buffer release
          * events are delivered after the buffer release queue and thread are
          * torn down.
+         *
+         * Do not destroy the WlEglStreamImages here because they may be
+         * in-flight and as soon as we unlock streamImagesMutex, the buffer
+         * release thread may access them.
          */
-        if (image->buffer) {
```
kbrenneman, line 2576: Probably not something for this change, but I think if we made `WlEglStreamImage` refcounted, then we could simplify this teardown dance quite a lot.

kbrenneman (approved):
I think this looks right.

PR #121, opened 2024-07-31
egl-wayland: Verify that errno is ETIME only on Linux

dkorkmazturk:
BSD variants do not define ETIME. So, the assert() call which verifies that errno is set to ETIME introduced in an earlier change could cause problems with debug builds on BSDs. This change adds preprocessor directives to limit that assertion to Linux.

### DIFF
--- a/src/wayland-eglsurface.c
+++ b/src/wayland-eglsurface.c
@@ -1206,7 +1206,9 @@ wlEglSurfaceCheckReleasePoints(WlEglDisplay *display, WlEglSurface *surface)
                            DRM_SYNCOBJ_WAIT_FLAGS_WAIT_AVAILABLE,
                            &firstSignaled) != 0) {
         /* A timeout is the only type of error we expect here */
+#ifdef ETIME
         assert(errno == ETIME);
+#endif
         goto end;
     }
 


src/wayland-eglsurface.c
```diff
@@ -1206,7 +1206,9 @@ wlEglSurfaceCheckReleasePoints(WlEglDisplay *display, WlEglSurface *surface)
                            DRM_SYNCOBJ_WAIT_FLAGS_WAIT_AVAILABLE,
                            &firstSignaled) != 0) {
         /* A timeout is the only type of error we expect here */
+#ifdef __linux__
```
kbrenneman, line 1209: Rather than `#ifdef __linux__`, we should just check `#ifdef ETIME` directly.

dkorkmazturk: Updated

kbrenneman (approved):
Looks good to me.

amshafer:
Fwiw it looks like FreeBSD uses `ETIMEDOUT`, that's what FreeBSD's DRM defines ETIME to. Since this is just an assert I think we are fine with what was merged here.

PR #122, opened 2024-07-31
egl-wayland: Add __NV_DISABLE_EXPLICIT_SYNC environment variable

amshafer:
There are still many application bugs caused by incorrect threading guarantees when running with explicit sync. This change implements a method for users to disable explicit sync manually for a particular application, in case there are issues with explicit sync.

Setting __NV_DISABLE_EXPLICIT_SYNC=1 will prevent the usage of explicit sync and fall back to the old method.

Tested disabling explicit sync on KDE, built on FreeBSD as well.


dkorkmazturk:
This looks good to me.

Issue #118, opened 2024-07-19
OBS Studio began to crash after egl-wayland 1.1.14 was released


LibertyGM:
If I open the "Windowed projector" in OBS Studio, then OBS crashes. If you roll back to version 1.1.13, then everything works.

For more information, see the video

warning: The Wayland connection experienced a fatal error: Protocol error

[20240719_072105.webm](https://github.com/user-attachments/assets/3cef4244-be5c-4df8-ba48-cc30cff9f7fa)

LibertyGM:
Operating System: Arch Linux 
KDE Plasma Version: 6.1.3
KDE Frameworks Version: 6.4.0
Qt Version: 6.7.2
Kernel Version: 6.10.0-arch1-1 (64-bit)
Graphics Platform: Wayland
Processors: 32 × Intel® Core™ i9-14900KF
Memory: 62.6 ГиБ of RAM
Graphics Processor: NVIDIA GeForce RTX 4080/PCIe/SSE2
Manufacturer: ASUS

thesword53:
I have the same issue with Firefox

LibertyGM:
In general, my whole system began to work unstable. If you go back to the old version, then everything is OK.

joshrandall8478:
I started having this issue as well

ptr1337:
> I have the same issue with Firefox

Firefox is not related to that. This was a firefox issue, which has been partially solved in the nightly.
It seems there is an outstanding issue in mesa too.

thesword53:
> > I have the same issue with Firefox
> 
> Firefox is not related to that. This was a firefox issue, which has been partially solved in the nightly. It seems there is an outstanding issue in mesa too.

I only have the issue with `egl-wayland` 1.1.14. With the 1.1.13 version I don't have any Firefox crash.

MadByteDE:
> > > I have the same issue with Firefox
> > 
> > 
> > Firefox is not related to that. This was a firefox issue, which has been partially solved in the nightly. It seems there is an outstanding issue in mesa too.
> 
> I only have the issue with `egl-wayland` 1.1.14. With the 1.1.13 version I don't have any Firefox crash.

Same for me with firefox. Also might have something to do with the 555 driver and kwin, didn't read to much into it, but when it happens I get:

```log
[...]

Issue #117, opened 2024-07-18
Version 1.1.14 is killing my firefox and brave


dmaliog:
Version 1.1.14 is killing my firefox and brave.
 firefox no longer works and in brave the title disappears when maximizing

⋊> ~ on master ⨯ inxi -G                                                                                                                             04:16:42
Graphics:
  Device-1: NVIDIA GP102 [GeForce GTX 1080 Ti] driver: nvidia v: 555.58.02
  Display: wayland server: X.org v: 1.21.1.13 with: Xwayland v: 24.1.1
    compositor: kwin_wayland driver: X: loaded: nvidia gpu: nvidia
    resolution: 2560x1440
  API: EGL v: 1.5 drivers: nvidia,swrast,zink
    platforms: gbm,wayland,x11,surfaceless,device
  API: OpenGL v: 4.6.0 compat-v: 4.5 vendor: nvidia mesa v: 555.58.02
    renderer: NVIDIA GeForce GTX 1080 Ti/PCIe/SSE2
  API: Vulkan v: 1.3.279 dri

OS: EndeavourOS Linux x86_64 
Host: B550 GAMING X V2 
Kernel: 6.10.0-1-mainline 
Uptime: 13 mins 
Packages: 1792 (pacman), 4 (flatpak), 4 (snap) 
Shell: fish 3.7.1 
Resolution: 2560x1440 
DE: Plasma 6.1.2 
WM: kwin 
Theme: Breeze-Dark [GTK2], Breeze [GTK3] 
Icons: Papirus-Dark [GTK2/3] 
Terminal: konsole 
CPU: AMD Ryzen 5 5600 (12) @ 3.500GHz 
GPU: NVIDIA GeForce GTX 1080 Ti 
Memory: 3686MiB / 32000MiB 

I had to roll back...

ionenwks:
Was trying to reproduce this in Gentoo given we've barely just added 1.1.14, but firefox-128.0 was working fine w/ plasma 6. Turns out it's because we do MOZ_ENABLE_WAYLAND=1 by default (aka native wayland), issues only manifest when that is disabled (=0) and it uses xwayland.

Output looks like this when disabled:
```
$ MOZ_ENABLE_WAYLAND=0 firefox
[GFX1-]: Wayland protocol error: wp_linux_drm_syncobj_surface_v1@67: error 4: explicit sync is used, but no acquire point is set

[...]

Issue #110, opened 2024-05-22
Application crashes when using explicit sync


Molytho:
Reference: https://github.com/NVIDIA/egl-wayland/pull/104#issuecomment-2123897621

While I don't use KDE I have similar issues with sway although the only application I encountered such crashes where firefox and thunderbird.

The issue here is a race between sending requests to the compositor:

> [3880880.548]  -> wl_compositor@36.create_region(new id wl_region@83)
> [3880880.573]  -> wl_region@83.add(0, 0, 1916, 1078)
> [3880880.584]  -> wl_surface@61.set_opaque_region(wl_region@83)
> [3880880.591]  -> wl_region@83.destroy()
> [3880880.603]  -> wl_surface@61.attach(wl_buffer@71, 0, 0)
> [3880880.631]  -> wl_surface@61.commit()
> [3880880.687]  -> wp_linux_drm_syncobj_surface_v1@74.set_acquire_point(wp_linux_drm_syncobj_timeline_v1@53, 0, 8)
> [3880880.698]  -> wp_linux_drm_syncobj_surface_v1@74.set_release_point(wp_linux_drm_syncobj_timeline_v1@82, 0, 2)
> [3880880.704]  -> wl_surface@61.damage(0, 0, 1916, 1078)
> [3880880.712]  -> wl_surface@61.commit()
> [3880880.732]  -> wl_display@1.sync(new id wl_callback@79)

[full log](https://github.com/NVIDIA/egl-wayland/files/15403635/wayland_debug_full.log)

This is a portion of the WAYLAND_DEBUG log when such a crash occurs (from thunderbird).
What I imagine happens here is that one thread calls set_opaque_region on the surface while another attaches a new buffer.
The commit after the set_opaque_region call happens to be in between attaching the buffer and setting the corresponding syncobj points. This is per definition a protocol violation of wp_linux_drm_syncobj_surface_v1.

So the real issue is that there is no way to send multiple request (like attach, set_acquire_point, set_release_point) atomically to the compositor.

MaLoLHD:
I have also seen this issue on GNOME. I've also used WAYLAND_DEBUG to get the log for this, here are the last few lines:

> [3561462.592]  -> wl_compositor@32.create_region(new id wl_region@77)
[3561462.607]  -> wl_region@77.add(0, 0, 1920, 1048)
[3561462.613]  -> wl_surface@63.set_opaque_region(wl_region@77)
[...]

PR #123, opened 2024-08-02
egl-wayland: Send explicit sync points before attaching a surface

amshafer:
Currently we call `wl_surface_attach`, then `send_explicit_sync_points`. This can theoretically cause issues due to `send_explicit_sync_points` exiting due to an error, causing `wlEglSendDamageEvent` to exit without finishing the surface commit. Not only is `wl_surface_commit` never called, but we are missing explicit sync points from the commit which is a protocol error.

This change switches the order, calling `send_explicit_sync_points` first. This is valid because the `linux_drm_syncobj_v1` protocol applies the acquire/release points to the attached buffer at commit time.

This affects at least some of the crashes seen in #110, #117  and #118

### DIFF
--- a/src/wayland-eglsurface.c
+++ b/src/wayland-eglsurface.c
@@ -266,19 +266,24 @@ wlEglSendDamageEvent(WlEglSurface *surface, struct wl_event_queue *queue)
             image->attached = EGL_TRUE;
         }
 
-        wl_surface_attach(surface->wlSurface,
-                          surface->ctx.currentBuffer,
-                          surface->dx,
-                          surface->dy);
-
         /*
          * Send our explicit sync acquire and release points. This needs to be done
          * as part of the surface attach as it is a protocol error to specify these
          * points without attaching a buffer in the same commit.
+         *
+         * Perform this before wl_surface_attach in case there is an error importing
+         * the syncfd at the current timeline point. If this errors out after the
+         * attach has happened then we are stuck with a protocol error from not
+         * specifying the timeline sync points.
          */
         if (!send_explicit_sync_points(surface->wlEglDpy, surface, image)) {
             return EGL_FALSE;
         }
+
+        wl_surface_attach(surface->wlSurface,
+                          surface->ctx.currentBuffer,
+                          surface->dx,
+                          surface->dy);
     }
 
     wl_surface_damage(surface->wlSurface, 0, 0,


amshafer:
CC: @mahkoh for pointing this out (thanks!)

cubanismo (approved):

furegulo:
I installed this patch on my Arch-based CachyOS system, with Nvidia 4070ti Super. I'm still getting a few random crashes with Firefox-based browsers with error message 

> cachyos-x8664 cachy-browser[12103]: Exiting due to channel error. 

Though this seems a lot more stable than the previous egl-wayland, which Arch reverted to the older one.

VictorR2007:
> I'm still getting a few random crashes with Firefox-based browsers with error message

Firefox Nightly is actively fixing such errors. With each update, Firefox Nightly reduces the number of Firefox crashes.

entailz:
Still crashing here for me as well on all branches of Firefox including Dev and Nightly, although like VictorR2007 said above, the crashes are less frequent.
```
Crash Annotation GraphicsCriticalError: |[0][GFX1-]: Wayland protocol error: wp_linux_drm_syncobj_surface_v1#74: error 5: Missing timeline
 (t=3.34461) [GFX1-]: Wayland protocol error: wp_linux_drm_syncobj_surface_v1#74: error 5: Missing timeline

Gdk-Message: 11:29:39.714: Error 71 (Protocol error) dispatching to Wayland display.
Exiting due to channel error.
Exiting due to channel error.
Crash Annotation GraphicsCriticalError: |[C0][GFX1-]: CompositorBridgeChild receives IPC close with reason=AbnormalShutdown (t=1.65293) Exiting due to channel error.
```

amshafer:
@entailz Can you please include a log of that crash with `WAYLAND_DEBUG=1`? Also would be helpful to know what compositor.

entailz:
I'm on Hyprland with explicit sync enabled, here's a log from Firefox ESR
[logs.txt](https://github.com/user-attachments/files/16499604/logs.txt)

entailz:
I just tested this commit on 560.28.03 with explicit sync enabled and I'm still getting the same crash in Firefox 
```
 Wayland protocol error: wp_linux_drm_syncobj_surface_v1#72: error 5: Missing timeline
```

amshafer:
You might want to check with the Hyprland devs if you haven't already, from your log it looks like the correct commit sequences are happening but according to the new Hyprland linux-drm-syncobj implementation the release timeline/point was not specified. It might be worth seeing if they find anything obviously wrong with their timeline tracking, just to verify.

I'm also looking into a potential issue with egl-wayland's event queue usage. There's a chance it could be causing what you see but I haven't reproduced your crash myself so it's hard to say. I'll get back to you about testing that since you have a more consistent repro for some reason.

entailz:
Okay, will do. Thanks for taking a look at the logs. I was told it could potentially be a Firefox issue, does that sound right? It seems to happen less on firefox nightly.

vaxerski:
hi! hyprland dev here. This is an nvidia bug, this didn't fix this one edge case :(

took me a while to spot this one, but:

20240805_23h42m06s

mahkoh:
If at commit time a syncobj_surface exists, then the content update must contain acquire and release points if and only if it attaches a non-null buffer.

vaxerski:
@entailz I've moved the check back after the buffer check in https://github.com/hyprwm/Hyprland/pull/7151/commits/b0437c9f76dc3533eed23f5a9aec186d4a32f4d3

If I am reading right, this now should error out with `WP_LINUX_DRM_SYNCOBJ_SURFACE_V1_ERROR_NO_BUFFER` because timeline 61 is attached while there is no buffer

mahkoh:
>ah, right, my bad, buffer 65 is destroyed. You're right

The buffer being destroyed is irrelevant. Destroying a buffer has no effect on surfaces it is attached to.

>Though... 61 is still attached and the buffer is gone... so that's still wrong no?

The commit your last arrow points to does not interact with the syncobj protocol at all because it does not attach a buffer. Its only contents are a frame request.

vaxerski:
makes sense I guess, thanks!

entailz:
@vaxerski working now, no crashes! thanks both of you for your patience.

PR #124, opened 2024-08-06
egl-wayland: Fix eglCreateSync and streamAcquireImage handling

amshafer:
This fixes two issues:
* eglCreateSync takes ownership of the fd passed to it, so we shoul not close the fd after passing it in.
* eglStreamAcquireImage's failure case only bails out of the function if acquireSync is valid, which is not the case when explicit sync is not in use.

### DIFF
--- a/src/wayland-egldisplay.c
+++ b/src/wayland-egldisplay.c
@@ -1261,7 +1261,6 @@ static void wlEglCheckDriverSyncSupport(WlEglDisplay *display)
     attribs[4] = EGL_NONE;
     eglSync = display->data->egl.createSync(dpy, EGL_SYNC_NATIVE_FENCE_ANDROID,
                                             attribs);
-    close (syncFd);
 
     /* If the call failed then the driver version is recent enough */
     if (eglSync == EGL_NO_SYNC_KHR &&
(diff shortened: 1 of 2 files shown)


amshafer:
CC: @mahkoh

dkorkmazturk:
Thanks for working on this @amshafer, LGTM.

scaronni:
@amshafer please merge, it is also confirmed to fix the Firefox 128 crashes on Fedora: https://bodhi.fedoraproject.org/updates/FEDORA-2024-3a273427fc

Nodens-:
@amshafer I can confirm that with git egl-wayland and both PR123 plus PR124 merged, I have had no fd crashes for 24 hours on 555 driver. Waiting 560.31.02 release on rpmfusion so I can test with that version too.

ofourdan:
Is that supposed to work with the r555 driver as well?

I am using this code on Fedora 40 (from @scaronni 's https://bodhi.fedoraproject.org/updates/FEDORA-2024-3a273427fc) with the NVIDIA driver version 555.58.02 and I am still seeing a suspectful high rate of crashes in Firefox 128.0.3 when playing Youtube.

ofourdan:
Ah nevermind, looks like I failed at picking up the update yesterday, it works much better now! Sorry for the noise.

Nodens-:
Bad fd crashes not reproducible anymore with 560 branch either

scaronni:
> Ah nevermind, looks like I failed at picking up the update yesterday, it works much better now! Sorry for the noise.

Will update to 1.1.15 in the same update as soon as it's tagged.

PR #138, opened 2024-08-23
Provide damage rectangles to wl_surface to avoid damaging entire surface when only parts of it need to be

Gert-dev:
This is a rebase of a single commit from #50 that provides damage rectangles for wl_surfaces so the compositor knows which parts of the surfaces are damaged instead of just always damaging the surface as a whole.

I set @chergert as commit author since the changes of this commit are identical to his.

I didn't include the part that supports the damage_thread yet from #50 because [the damage thread is apparently gone when using explicit sync](https://github.com/NVIDIA/egl-wayland/pull/50#issuecomment-2305484091), this seems to also be sufficient to get applications such as Nautilus to only redraw the parts they need to, and this keeps the diff smaller. (In https://github.com/NVIDIA/egl-wayland/pull/50#issuecomment-2302748546 I also rebased the other commit, but we can add that separately if it still turns out to be useful.)

### DIFF
--- a/src/wayland-eglsurface.c
+++ b/src/wayland-eglsurface.c
@@ -242,9 +242,13 @@ send_explicit_sync_points (WlEglDisplay *display, WlEglSurface *surface,
 }
 
 EGLBoolean
-wlEglSendDamageEvent(WlEglSurface *surface, struct wl_event_queue *queue)
+wlEglSendDamageEvent(WlEglSurface *surface,
+                     struct wl_event_queue *queue,
+                     EGLint *rects,
+                     EGLint n_rects)
 {
     struct wl_display *wlDpy = surface->wlEglDpy->nativeDpy;
+    EGLint i;
 
     if (surface->ctx.wlStreamResource) {
         /* Attach same buffer to indicate new content for the surface is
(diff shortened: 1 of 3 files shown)


src/wayland-eglsurface.c
```diff
@@ -286,8 +290,22 @@ wlEglSendDamageEvent(WlEglSurface *surface, struct wl_event_queue *queue)
...
+    if (n_rects > 0 &&
+        (wl_proxy_get_version((struct wl_proxy *)surface->wlSurface) >=
+         WL_SURFACE_DAMAGE_BUFFER_SINCE_VERSION)) {
+        for (i = 0; i < n_rects; i++) {
+            int y = rects[i*4+1] + rects[i*4+3];
+            wl_surface_damage_buffer(surface->wlSurface,
+                                     rects[i*4],
+                                     surface->height - y,
+                                     rects[i*4+2],
+                                     rects[i*4+3]);
```
amshafer, line 302: Thanks I think this looks pretty good, just a couple nitpicks:

I think we should have a small comment here mentioning that the coordinate systems are flipped between `eglSwapBuffersWithDamage()` and `wl_surface_damage_buffer()`.  Also naming this variable `y` seems slightly confusing, as technically `rects[i*4+1]` is really the passed-in y position and `int y` here is actually the ending point of the damage rectangle. Maybe naming it something like `max_y` would be more accurate?

We have a lot of `i*4` here, I think it might be more readable if we created a `rect_offset` variable to hold this? i.e. `rects[rect_offset+2]`.

kbrenneman: > We have a lot of i*4 here, I think it might be more readable if we created a rect_offset

For that, it would probably be more readable to just declare a second pointer so that we can index from zero:
```C
const EGLint *rect = &rects[i * 4];
int inv_y = surface->height - rect[1] + rect[3];
wl_surface_damage_buffer(surface->wlSurface, rect[0], inv_y, rect[2], rect[3]);
```

Gert-dev: Sounds good. I applied the suggestion by @kbrenneman and added a small comment about the Y being flipped.

amshafer (approved):

kbrenneman (approved):

src/wayland-eglsurface.c
```diff
@@ -286,8 +290,21 @@ wlEglSendDamageEvent(WlEglSurface *surface, struct wl_event_queue *queue)
...
+         WL_SURFACE_DAMAGE_BUFFER_SINCE_VERSION)) {
+        for (i = 0; i < n_rects; i++) {
+            const EGLint *rect = &rects[i * 4];
+            // Coordinate systems are flipped between eglSwapBuffersWithDamage
+            // and wl_surface_damage_buffer, so invert Y values.
+            int inv_y = surface->height - (rect[1] + rect[3]);
+            wl_surface_damage_buffer(surface->wlSurface, rect[0], inv_y, rect[2], rect[3]);
+        }
+    } else {
+        wl_surface_damage(surface->wlSurface, 0, 0, UINT32_MAX, UINT32_MAX);
```
Molytho, line 304: The width and height parameters of wl_surface.damage are of type int so this should bi INT32_MAX

kbrenneman: Ah, good point. I'll get that fixed.

Issue #140, opened 2024-08-26
Visual glitches in KDE 6.1 desktop in 16-1.1


CodionCA:
### Issue description:
Immediately after upgrading to 16-1.1 from 13-1.1 I experienced wallpapers not loading on either display intermittently, the task bar being cut off, any video playback being 1/4 of the framerate etc. Downgrading to the previous version I was on 13-1.1 fixed the issues after a reboot.

### Evidence
![Image](https://github.com/user-attachments/assets/9ab4d0f1-4d70-4ecf-ba52-2acbf2fe26f3)
This happens after opening a program from the task tray then closing it from the window itself.

### Extra info
OS: EndeavorOS
Nvidia Driver: 560.35.03
EGL-Wayland: 4:1.1.16-1.1
KDE Plasma: 6.1.4

ddxtanx:
I'm getting extremely similar artifacts on sway (1.10-dev) with egl-wayland-git-1.1.16.r1.g2d31baa specifically. Things like animated backgrounds only update when a different window renders an update (e.g. I interact with a terminal) and otherwise my mouse leaves a tail of updated pixels when I drag it to different areas. This issue did not appear in egl-wayland-git-1.1.16.r0.g5c5a93c.

PR #141, opened 2024-08-27
Use INT32_MAX for wl_surface_damage

kbrenneman:
This is a follow-up to #138, to use INT32_MAX instead of UINT32_MAX for the wl_surface::damage request, since the protocol defines that as taking a signed int.


amshafer (approved):
This fixes #140

PR #144, opened 2024-09-06
egl-wayland: Accept device name from either wl_drm or linux_dmabuf

amshafer:
Commit 11386cc allowed for listening for the DRM device using the dmabuf protocol instead of wl_drm, when possible. Unfortunately some headless compositors (kwin) do not properly send the device id through dmabuf, even though they send the device path properly through wl_drm. This change relaxes our check to ensure that one of the protocols sends the information so that we can use it to verify compositor support on versions that behave like this.

This fixes some issues with automated testing on kwin.

### DIFF
--- a/src/wayland-egldisplay.c
+++ b/src/wayland-egldisplay.c
@@ -485,6 +485,7 @@ static void wl_drm_device(void *data, struct wl_drm *wl_drm, const char *name)
     WlServerProtocols *protocols = (WlServerProtocols *)data;
     (void) wl_drm;
 
+    free(protocols->drm_name);
     protocols->drm_name = strdup(name);
 }
 
@@ -575,6 +576,7 @@ dmabuf_feedback_check_done(void *data, struct zwp_linux_dmabuf_feedback_v1 *dmab
     assert(getDeviceFromDevId);
     if (getDeviceFromDevId(protocols->devId, 0, &drm_device) == 0) {
         if (drm_device->available_nodes & (1 << DRM_NODE_RENDER)) {
+            free(protocols->drm_name);
             protocols->drm_name = strdup(drm_device->nodes[DRM_NODE_RENDER]);
         }
 
@@ -632,6 +634,7 @@ registry_handle_global_check_protocols(
 
     if ((strcmp(interface, "wl_drm") == 0) && (version >= 2)) {
         protocols->wlDrm = wl_registry_bind(registry, name, &wl_drm_interface, 2);
+        wl_drm_add_listener(protocols->wlDrm, &drmListener, protocols);
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/wayland-egldisplay.c
```diff
@@ -828,10 +829,9 @@ static bool getServerProtocolsInfo(struct wl_display *nativeDpy,
                 wl_display_roundtrip_queue(nativeDpy, queue);
                 zwp_linux_dmabuf_feedback_v1_destroy(default_feedback);
             }
-        } else if (protocols->wlDrm) {
-            wl_drm_add_listener(protocols->wlDrm, &drmListener, protocols);
-            wl_display_roundtrip_queue(nativeDpy, queue);
         }
+
+        /* Check that one of our two protocols provided the device name */
         result = protocols->drm_name != NULL;
```
cubanismo, line 835: Do servers ever provide both protocols? I would have at least had to at some point to support a transition for clients. It looks like we'll leak at least one protocols->drm_name allocation in that case with this code.

kbrenneman:
Does kwin advertise `zwp_linux_dmabuf_v1` version 4+ in that configuration?

If so, then failing to send a `zwp_linux_dmabuf_feedback_v1::main_device` event is a clear spec violation.

src/wayland-egldisplay.c
amshafer: Yes they do, updated to free the existing drm_name before replacing it. This doesn't prioritize one source of the name over the other, it's just the last event to roll in. I think that should be fine, as these should be the same thing? This works with kwin since kwin sends a dev_t of zero through dmabuf and the path name through wl_drm, so we don't actually update the drm_name in our main_device handler as that devid doesn't exist. I think that should be fine?

amshafer:
Yes kwin does advertise 4+, and it does send the main_device event. The issue is it sends a devid of zero which we can't get a name from and therefore don't update the drm path. Technically kwin isn't violating the spec in any way, as I can't find anything prohibiting zero from being a valid dev_t value? I doubt this is wanted behavior from kwin though since they do advertise the correct path in wl_drm. Even so, this fix is still needed so that we can test older kwins that behave like this even if they advertise the real dev_t in the future.

dkorkmazturk:
I remember running into this on the Vulkan side. In that case, we just end up falling back to PRIME path since the main device advertised by the compositor does not match the device Vulkan clients use. I didn't really look at how we handle PRIME render offload on EGL but would it be possible to follow a similar approach here?

amshafer:
We already check the output of `checkNvidiaDrmDevice` and avoid continuing without a NVIDIA-based server unless PRIME is configured by the user, so I don't think we can do that here.

kbrenneman:
> Yes kwin does advertise 4+, and it does send the main_device event. The issue is it sends a devid of zero which we can't get a name from and therefore don't update the drm path. Technically kwin isn't violating the spec in any way, as I can't find anything prohibiting zero from being a valid dev_t value? I doubt this is wanted behavior from kwin though since they do advertise the correct path in wl_drm. Even so, this fix is still needed so that we can test older kwins that behave like this even if they advertise the real dev_t in the future.

I don't remember if zero is a valid major/minor device number, but if it doesn't correspond to an actual device node, then it isn't a valid device. It would be worth filing a bug against kwin if there isn't already.

src/wayland-egldisplay.c
cubanismo: Yeah, I imagine it's fine to fix the buggy case. I was just curious what happens in the regular case where a valid dev_t is sent. I suppose they'd generally be the same, but it seems like we would want to prefer the dmabuf one if it were ever the case that they didn't match. This definitely seems like an improvement either way.

cubanismo (approved):

PR #145, opened 2024-09-09
egl-wayland: The `image` parameter of `send_explicit_sync_points` should not be NULL.

shiningdracon:
The `send_explicit_sync_points` function assumes that none of its three parameters are NULL and does not perform NULL pointer checks. However, during runtime, the `image` parameter can potentially be NULL, leading to a crash.

Attached crash log.

[crash_log.txt](https://github.com/user-attachments/files/16924811/crash_log.txt)

### DIFF
--- a/src/wayland-eglsurface.c
+++ b/src/wayland-eglsurface.c
@@ -268,6 +268,8 @@ wlEglSendDamageEvent(WlEglSurface *surface,
         if (image) {
             surface->ctx.currentBuffer = image->buffer;
             image->attached = EGL_TRUE;
+        } else {
+            return EGL_FALSE;
         }
 
         /*


amshafer:
Thanks, what application does this happen with? This sounds like an issue of us not ensuring an image is available, so I'd like to narrow down the root cause of that.

amshafer:
Can you please provide details to reproduce? What kwin version, did this happen on startup or triggered by behavior, etc?

shiningdracon:
I’m not sure how to reproduce the issue. I left my PC on overnight, and when I checked it the next morning, five crash logs had appeared. However, the problem hasn’t occurred again since then.

amshafer:
Sorry for the delay, I've been trying to reproduce this so I can figure out exactly why the image tracking ends up in this state but haven't observed the crash at all. I agree we do need to handle this. Could you please rebase this? Change itself seems good.

+CC: @cubanismo to take a second look

amshafer:
Implemented in #153

PR #146, opened 2024-09-17
egl-wayland: fix device name case where only wl_drm exists

amshafer:
This fixes a problem in commit 6355c16 where we do not properly wait for a roundtrip to get the wl_drm name, which can lead to failing to initialize and falling back to another ICD. This fix also properly prioritizes the device provided by dmabuf feedback over the wl_drm name, only using wl_drm if dmabuf did not provide anything.

### DIFF
--- a/src/wayland-egldisplay.c
+++ b/src/wayland-egldisplay.c
@@ -634,7 +634,6 @@ registry_handle_global_check_protocols(
 
     if ((strcmp(interface, "wl_drm") == 0) && (version >= 2)) {
         protocols->wlDrm = wl_registry_bind(registry, name, &wl_drm_interface, 2);
-        wl_drm_add_listener(protocols->wlDrm, &drmListener, protocols);
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/wayland-egldisplay.c
```diff
@@ -833,6 +832,12 @@ static bool getServerProtocolsInfo(struct wl_display *nativeDpy,
             }
         }
 
+        /* If we didn't get a name through linux_dmabuf then fall back to wl_drm */
+        if (!protocols->drm_name && protocols->wlDrm) {
+            wl_drm_add_listener(protocols->wlDrm, &drmListener, protocols);
+            wl_display_roundtrip_queue(nativeDpy, queue);
+        }
```
mahkoh, line 839: Have you actually tested this? Surely at this point the wl_drm events have already been received and discarded by libwayland-client.

amshafer: Yes, this is actually restoring where this `wl_drm_add_listener` has been historically in egl-wayland, before 6355c16.

It works since the earlier roundtrip will bind the the protocols in the registry, but we need a second roundtrip to actually trigger handling of the wl_drm events the compositor sent. This is the problem this change fixes which my previous change had undone. iirc this double roundtrip pattern is present in other projects as well.

mahkoh: If zwp_linux_dmabuf_v1 is present, there are three roundtrips and the second one consumes the wl_drm events.

amshafer: Ah I see what you mean now, will fix.

PR #150, opened 2024-10-10
egl-wayland: enable CI with github actions

amshafer:
This enables CI for github actions builds, aimed for helping verify new pull requests. This adds tests for the meson and autotools builds on ubuntu 24.04. Other distros are apparently not supported by default so we aren't able to expand testing this way.

We will need to add things like other distro support, FreeBSD CI, etc in the future but this is a good start for now.

### DIFF
--- a/.github/workflows/arch-build.yml
+++ b/.github/workflows/arch-build.yml
@@ -1,0 +1,16 @@
+name: Arch Build
+on: [push, pull_request]
+jobs:
+  Meson-Build:
+    runs-on: ubuntu-24.04
+    container:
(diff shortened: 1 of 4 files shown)


scaronni:
We're doing it in containers with DKMS: https://github.com/dell/dkms/blob/master/.github/workflows/tests.yml

Maybe this helps.

amshafer:
Thanks, that's helpful. Updated to test LLVM, fix dependencies, used containers for arch, in-person feedback from @dkorkmazturk

cubanismo (approved):

PR #151, opened 2024-10-13
egl-wayland: Fix use after free at library teardown

Molytho:
This patch extends the hack described [here](https://github.com/NVIDIA/egl-wayland/blob/c10c5300483a8ec975e64e5d76c0fb00ac94e026/src/wayland-egldisplay.c#L692) to wlEglDestroyFeedback which might destroy a zwp_linux_dmabuf_feedback_v1 object.

I encountered a segmentation fault at termination of the game celeste because of this

### DIFF
--- a/src/wayland-egldisplay.c
+++ b/src/wayland-egldisplay.c
@@ -729,10 +729,10 @@ static EGLBoolean terminateDisplay(WlEglDisplay *display, EGLBoolean globalTeard
      * destroy the display connection itself */
     wlEglDestroyAllSurfaces(display);
 
-    wlEglDestroyFormatSet(&display->formatSet);
-    wlEglDestroyFeedback(&display->defaultFeedback);
-
     if (!globalTeardown || display->ownNativeDpy) {
+        wlEglDestroyFormatSet(&display->formatSet);
+        wlEglDestroyFeedback(&display->defaultFeedback);
+
         if (display->wlRegistry) {
             wl_registry_destroy(display->wlRegistry);
             display->wlRegistry = NULL;


src/wayland-egldisplay.c
```diff
@@ -730,7 +730,7 @@ static EGLBoolean terminateDisplay(WlEglDisplay *display, EGLBoolean globalTeard
     wlEglDestroyAllSurfaces(display);
 
     wlEglDestroyFormatSet(&display->formatSet);
```
amshafer, line 732: nit: Can you please remove this call to `wlEglDestroyFormatSet`? It predates your change but it's not needed as  `wlEglDestroyFeedback` will reset the tranches and call this anyway.

src/wayland-egldisplay.c
```diff
@@ -730,7 +730,7 @@ static EGLBoolean terminateDisplay(WlEglDisplay *display, EGLBoolean globalTeard
...
-    wlEglDestroyFeedback(&display->defaultFeedback);
+    wlEglDestroyFeedback(&display->defaultFeedback, globalTeardown);
```
amshafer, line 733: I actually think it might make more sense to move this call down into the `if (!globalTeardown || display->ownNativeDpy) {..}` block below this, just above the `if (display->wlDmaBuf) ` part. That would prevent us from having to change `wlEglDestroyFeedback` at all and also respect the `display->ownNativeDpy` checking we do. As-is I think we would defer `wlEglDestroyFeedback` but destroy everything else.

Molytho: I did it like that because wlEglDestroyFeedback also has a munmap call which isn't called this way but as library unload usually means application shutdown anyway I guess it's reasonable to ignore this case.

src/wayland-egldisplay.c
Molytho: Are you sure about this? To me it looks like this is a completely independent format set. It is used if the dmabuf version is <4.

src/wayland-egldisplay.c
amshafer: Well right now since it isn't checked like you said it's possible for the `munmap` to happen twice if `terminateDisplay` is called multiple times for the deferred free workaround, so that's another problem fixed by this suggestion. If we move `wlEglDestroyFeedback` into the guarded `globalTeardown` section then we avoid this as well.

src/wayland-egldisplay.c
amshafer: Ah good point. I think this should be moved to stay next to `wlEglDestroyFeedback` for consistency.

amshafer:
++@dkorkmazturk to take a look at this as well

dkorkmazturk:
Looks good to me. Thank you for fixing this issue.

amshafer (approved):
Can you please rebase this to pick up the CI that was recently added?

PR #155, opened 2024-11-15
Check duplicate native surfaces

kbrenneman:
By the EGL spec, you're not allowed to create more than one EGLSurface from the same native window/pixmap. Currently, egl-wayland only partially checks for that, but you can still cause problems in a slightly more roundabout way.

While egl-wayland does check if the app tries to create a second EGLSurface from the same `wl_egl_window`, it's still possible to create more than one EGLSurface from the same `wl_surface`. Since `wl_egl_window_create` itself doesn't check for duplicates, you can create more than one `wl_egl_window` from the same `wl_surface`, and then create an EGLSurface for each.

With this change, egl-wayland will scan the list of EGLSurfaces to check if any of them use the same underlying `wl_surface`, and if it finds one, it'll fail eglCreateWindowSurface with EGL_BAD_ALLOC.

I also fixed a segfault that it would run into in early-failure cases like this one, where it hasn't allocated the `WlEglSurface` yet, but the cleanup code tries to dereference it.

### DIFF
--- a/src/wayland-eglsurface.c
+++ b/src/wayland-eglsurface.c
@@ -2652,7 +2652,10 @@ EGLSurface wlEglCreatePlatformWindowSurfaceHook(EGLDisplay dpy,
     WlEglDisplay         *display = wlEglAcquireDisplay(dpy);
     WlEglPlatformData    *data    = NULL;
     WlEglSurface         *surface = NULL;
+    WlEglSurface         *existingSurf = NULL;
     struct wl_egl_window *window  = (struct wl_egl_window *)nativeWin;
+    struct wl_surface    *wsurf   = NULL;
+    long int              wver    = 0;
     EGLBoolean            res     = EGL_FALSE;
     EGLint                err     = EGL_SUCCESS;
     EGLint                surfType;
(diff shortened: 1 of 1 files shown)


src/wayland-eglsurface.c
```diff
@@ -2683,6 +2685,23 @@ EGLSurface wlEglCreatePlatformWindowSurfaceHook(EGLDisplay dpy,
         goto fail;
     }
 
+    getWlEglWindowVersionAndSurface(window, &wver, &wsurf);
+    if (wsurf == NULL) {
+        err = EGL_BAD_ALLOC;
+        goto fail;
+    }
```
amshafer, line 2692: Here we store our wl_surface in `wsurf` so we don't prematurely set it, but doesn't this leak the surface if we jump to `fail` later on (like in line 2708? Before this change I think this would get cleaned up as part of `wlEglDestroySurface` but now this isn't recorded in `surface` until later?

I think we can go ahead and set `surface->wlSurface` here? Afaict that shouldn't affect anything between here and 2779 and properly handles the surface cleanup.

kbrenneman: `getWlEglWindowVersionAndSurface` doesn't allocate anything, it just figures out the correct offset for the `wl_surface` pointer in the `wl_egl_surface` struct.

That said, using the same `surface` variable in the loop *would* cause problems in the failure path, because it'll try to clean up an existing `WlEglSurface`.

amshafer: Ah good point, thanks that looks better now.

amshafer (approved):

PR #160, opened 2025-01-11
Fix read return value check in buffer_release_thread

al42and:
Previously, we were reading 0 bytes and were passing uninitialized value to `switch`. Likely, not much changed in practice (all codepaths were just terminating the thread, regardless of the success of the `read`), but now we don't trigger undefined behavior.


amshafer (approved):
Thanks for finding this, looks good to me.

PR #168, opened 2025-04-03
egl-wayland: Add support for tegradisp-drm

psruthikNV:
This change adds support in the Wayland EGL Platform library for the tegradisp-drm DRM driver meant for Automotive (DRIVE OS) Tegra platforms.


amshafer:
Change itself seems reasonable to me, in line with other tegra names we have used.

Can you please provide details (either here or through email if a public forum isn't the right place) about the new tegradisp-drm platform and how it differs from the others? I'd like to know about what egl-wayland is going to be running on before supporting something, mostly so I can better triage future bug reports.

cubanismo (approved):

psruthikNV:
Looks like I don't have merge rights for this project.

@amshafer If everything with this PR looks good, can you help merge it as well? Thanks!

PR #169, opened 2025-04-03
Fix reporting EGL_EXT_present_opaque

kbrenneman:
This removes `EGL_EXT_present_opaque` from the `EGL_EXT_PLATFORM_DISPLAY_EXTENSIONS` query, and instead adds a hook for `eglQueryString` to report it.

Any extensions in `EGL_EXT_PLATFORM_DISPLAY_EXTENSIONS` get added to *every* EGLDisplay, not just the ones that belong to egl-wayland. That's what we want for things like `EGL_WL_bind_wayland_display`, which is implemented in egl-wayland, but used in a compositor (generally with a device or GBM-based display).

For `EGL_EXT_present_opaque`, we need to report that through an `eglQueryString` hook instead. Unfortunately, current drivers won't call an eglQueryString hook, so without an updated driver, this will remove `EGL_EXT_present_opaque` from Wayland EGLDisplays as well. But, that's still better than advertising it on non-Wayland EGLDisplays where it doesn't work.

### DIFF
--- a/include/wayland-egldisplay.h
+++ b/include/wayland-egldisplay.h
@@ -182,6 +182,8 @@ typedef struct WlEglDisplayRec {
     WlEglDmaBufFeedback defaultFeedback;
 
     EGLBoolean primeRenderOffload;
+
+    char *extensionString;
 } WlEglDisplay;
 
 typedef struct WlEventQueueRec {
--- a/include/wayland-egldisplay.h
+++ b/include/wayland-egldisplay.h
@@ -223,6 +223,7 @@ EGLBoolean wlEglGetConfigAttribHook(EGLDisplay dpy,
 EGLBoolean wlEglQueryDisplayAttribHook(EGLDisplay dpy,
                                        EGLint name,
                                        EGLAttrib *value);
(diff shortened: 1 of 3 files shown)


amshafer (commented):
Concept looks good.

> Unfortunately, current drivers won't call an eglQueryString hook, so without an updated driver, this will remove EGL_EXT_present_opaque from Wayland EGLDisplays as well. But, that's still better than advertising it on non-Wayland EGLDisplays where it doesn't work.

Sorry I think I'm missing something, why do current drivers not call `eglQueryString` hooks? I agree that we need to stop advertising `EGL_EXT_present_opaque` in non egl-wayland displays but IMO rolling back support for an extension on all non-current drivers seems like a large impact. Is this all drivers prior to 570 or something? I worry we are trading a theoretical problem for a practical problem.

kbrenneman:
I fixed the indentation style, and I consolidated the two `strdup(PRESENT_OPAQUE_NAME)` into one place, which removed the need for the `if (len == 0)` check.

I'd also forgotten to check for a malloc failure.

amshafer (approved):
Ah okay I see what bug this is fixing now, makes sense. Thanks, updated changes LGTM.

PR #177, opened 2025-09-03
egl-wayland: Fix libdrm loading when wl_drm is not available

amshafer:
If the compositor does not support wl_drm and we fail to get the drmGetDeviceFromDevId symbol then the driver load will fail. This happens with steam due to it hijacking dlopen/dlsym. This change has us directly dlopen libdrm and get a symbol from it instead of relying on RTLD_DEFAULT.

### DIFF
--- a/src/wayland-egldisplay.c
+++ b/src/wayland-egldisplay.c
@@ -48,15 +48,14 @@ typedef struct WlServerProtocolsRec {
     struct zwp_linux_dmabuf_v1 *wlDmaBuf;
     dev_t devId;
 
+    WlEglPlatformData *pData;
+
     struct wl_drm *wlDrm;
     char *drm_name;
 } WlServerProtocols;
 
 /* TODO: Make global display lists hang off platform data */
 static struct wl_list wlEglDisplayList = WL_LIST_INITIALIZER(&wlEglDisplayList);
-
-static bool getDeviceFromDevIdInitialised = false;
-static int (*getDeviceFromDevId)(dev_t dev_id, uint32_t flags, drmDevice **device) = NULL;
 
 EGLBoolean wlEglIsWaylandDisplay(void *nativeDpy)
 {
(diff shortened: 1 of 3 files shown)


src/wayland-egldisplay.c
```diff
@@ -826,17 +825,12 @@ static bool getServerProtocolsInfo(struct wl_display *nativeDpy,
...
-            getDeviceFromDevIdInitialised = true;
-        }
-
         /*
          * if dmabuf feedback is available then use that. This will potentially
          * replace the drm_name provided by wl_drm, assuming the feedback provides
          * a valid dev_t.
          */
-        if (protocols->wlDmaBuf && getDeviceFromDevId) {
+        if (protocols->wlDmaBuf && protocols->pData->libdrmDlHandle && protocols->pData->getDeviceFromDevId) {
```
kbrenneman, line 833: Same as before, as long as we have `getDeviceFromDevId`, it doesn't matter if `libdrmDlHandle` is NULL.

src/wayland-eglhandle.c
```diff
@@ -151,6 +152,15 @@ wlEglCreatePlatformData(int apiMajor, int apiMinor, const EGLExtDriver *driver)
     res->callbacks.setError           = driver->setError;
     res->callbacks.streamSwapInterval = driver->streamSwapInterval;
 
+    /* try to find libdrm and drmGetDeviceFromDevId */
+    res->libdrmDlHandle = dlopen("libdrm.so.2", RTLD_LAZY);
```
kbrenneman, line 156: Minor nitpick, but it might be better to try RTLD_DEFAULT first, and then go with dlopen if that fails.

In most cases, the RTLD_DEFAULT case will succeed, so we'd avoid the extra overhead of calling dlsym.

Also, using RTLD_DEFAULT first means that you could still override `drmGetDeviceFromDevId` with an LD_PRELOAD library if you had some reason to do so.

Either way, we should add a comment here explaining why loading libdrm.so can give a different result than using RTLD_DEFAULT, since by the spec the two should be strictly equivalent.

kbrenneman (approved):
Looks good to me.

PR #178, opened 2025-10-01
egl-wayland: support linux-dmabuf-v6 

amshafer:
This enables support for dmabuf v6 and uses the new new set_sampling_device request to inform the compositor about what device the compositor should import this buffer on.

This also switches us from using the unstable to the stable linux-dmabuf protocol version.

https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/268

Starting as a draft, not yet fully tested.

### DIFF
--- a/.github/workflows/autoconf-build.yml
+++ b/.github/workflows/autoconf-build.yml
@@ -6,9 +6,17 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: sudo apt update
-      - run: sudo apt install -y wayland-protocols libdrm-dev libegl-dev
+      - run: sudo apt install -y libdrm-dev libegl-dev
       - run: sudo apt install -y libwayland-dev libwayland-egl-backend-dev eglexternalplatform-dev
       - run: sudo apt install -y meson ninja-build gcc
+      # Ubuntu's wayland-protocols is too old for linux-dmabuf v6.
+      - name: Build wayland-protocols 1.49 from source
+        run: |
+          curl -L https://gitlab.freedesktop.org/wayland/wayland-protocols/-/releases/1.49/downloads/wayland-protocols-1.49.tar.xz | tar xJ
+          cd wayland-protocols-1.49
+          meson setup build --prefix=/usr -Dtests=false
+          ninja -C build
+          sudo ninja -C build install
       - run: ./autogen.sh
       - run: make
       - run: sudo make install
--- a/.github/workflows/meson-build.yml
+++ b/.github/workflows/meson-build.yml
@@ -6,9 +6,17 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: sudo apt update
-      - run: sudo apt install -y wayland-protocols libdrm-dev libegl-dev
+      - run: sudo apt install -y libdrm-dev libegl-dev
       - run: sudo apt install -y libwayland-dev libwayland-egl-backend-dev eglexternalplatform-dev
       - run: sudo apt install -y meson ninja-build gcc
+      # Ubuntu's wayland-protocols is too old for linux-dmabuf v6.
+      - name: Build wayland-protocols 1.49 from source
+        run: |
+          curl -L https://gitlab.freedesktop.org/wayland/wayland-protocols/-/releases/1.49/downloads/wayland-protocols-1.49.tar.xz | tar xJ
+          cd wayland-protocols-1.49
+          meson setup build --prefix=/usr -Dtests=false
+          ninja -C build
+          sudo ninja -C build install
       - run: meson build
       - run: ninja -C build
       - run: sudo ninja -C build install
--- a/.github/workflows/meson-llvm-build.yml
+++ b/.github/workflows/meson-llvm-build.yml
@@ -6,9 +6,17 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: sudo apt update
-      - run: sudo apt install -y wayland-protocols libdrm-dev libegl-dev
+      - run: sudo apt install -y libdrm-dev libegl-dev
       - run: sudo apt install -y libwayland-dev libwayland-egl-backend-dev eglexternalplatform-dev
       - run: sudo apt install -y meson ninja-build clang
+      # Ubuntu's wayland-protocols is too old for linux-dmabuf v6.
+      - name: Build wayland-protocols 1.49 from source
+        run: |
+          curl -L https://gitlab.freedesktop.org/wayland/wayland-protocols/-/releases/1.49/downloads/wayland-protocols-1.49.tar.xz | tar xJ
+          cd wayland-protocols-1.49
+          meson setup build --prefix=/usr -Dtests=false
+          ninja -C build
+          sudo ninja -C build install
       - name: meson build
         run: meson build
         env:


Drakulix:
Note: While a valid implementation, this is specifically not how you should utilize the new functionality v6 provides.

The goal is to move any copy (if necessary) to the compositor rather than letting the client handle it. So if the compositor advertises any tranche with `Sampling`-flags for the nvidia gpu in question, egl-wayland should use the same code path as it does on single-gpu systems and set the nvidia gpu for `zwp_linux_buffer_params_v1_set_sampling_device`.

Only if the compositor cannot sample from the gpu directly, egl-wayland should consider the first tranche with the sampling flag as the main device and handle copying, using the gpu the buffer was imported to for `zwp_linux_buffer_params_v1_set_sampling_device`.

See also the wip mesa implementation for this protocol version: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/33344

amshafer:
Okay thanks that's good to confirm. I was intending to change it to something like what you described but haven't gotten to it yet. My main problem at the moment is testing this since the compositor implementations are in various WIP states, is the COSMIC/smithay implementation merged or is there an easy way I can test with it?

Drakulix:
> Okay thanks that's good to confirm. I was intending to change it to something like what you described but haven't gotten to it yet. My main problem at the moment is testing this since the compositor implementations are in various WIP states, is the COSMIC/smithay implementation merged or is there an easy way I can test with it?

Sorry for getting back to this rather late. There is a work-in-progress branch for cosmic-comp here, which I tested successfully with the previously linked mesa implementation: https://github.com/pop-os/cosmic-comp/pull/2513

Nothing merged yet, but hopefully soon! Compiling the branch should be pretty straight forward, but I am also planing to test the latest changes to this PR soon. The latest commit should now be able to skip the copy, right?

amshafer:
Yes it should be able to skip the copy, although I haven't finished properly testing it yet so keep that in mind. I can give the MR you linked a go and see if there are any issues that need ironing out.

fwiw we also plan on implementing this in the same manner for egl-wayland2 as that is the default on latest release branches > 580 (cc @kbrenneman)

src/wayland-egldisplay.c
```diff
@@ -1158,10 +1269,14 @@ EGLDisplay wlEglGetPlatformDisplayExport(void *data,
...
+    if (eglDevice == EGL_NO_DEVICE_EXT &&
+        (usePrimeRenderOffload || protocols.primeSamplingDev != 0)) {
         /*
-         * If __NV_PRIME_RENDER_OFFLOAD is set, then use an NVIDIA device. It
-         * doesn't matter which one.
+         * No NVIDIA EGL device matched the compositor's device.  Either the
+         * __NV_PRIME_RENDER_OFFLOAD env var was set explicitly, or the dmabuf
+         * feedback showed that the compositor cannot sample from any NVIDIA
+         * GPU directly (PRIME copy path required).  Pick the first available
+         * NVIDIA device for rendering.
```
Drakulix, line 1231: So how does this exactly work with v6 for EGL?

The checks further up in this function require either `__NV_PRIME_RENDER_OFFLOAD` to be set or an explicitly requested device. The latter won't be the case for the vast majority of egl-apps, which don't use `EGL_EXT_explicit_device`.

But then it always end up setting `primeSamplingDevce`, because this code treats `__NVIDIA_PRIME_RENDER_OFFLOAD` as explicitly requesting a prime-copy?

As a user how do I achieve the same, that `DRI_PRIME=1` would give me here with the wip mesa-PR? Which is no copy, but the nvidia device selected by default.

I assume for the vulkan case, there will always be a requested device, which means `__VK_LAYER_NV_optimus` is now required for explicit device selection or `MESA_VK_DEVICE_SELECT` (since that is a just a layer, it should work the same I assume?). That might be worthwhile to document somewhere?

Drakulix (commented):
Well this certainly seems to call `set_sampling_device` with the nvidia-gpu's dev_t on my hybrid laptop.

However I am not seeing the same performance improvements, I have previously seen with an AMD eGPU, though that setup was probably much more starved for bandwidth.

src/wayland-egldisplay.c
kbrenneman: Strictly speaking, `__NV_PRIME_RENDER_OFFLOAD` only affects whether we return NULL or non-NULL from `eglGetPlatformDisplay`.

For EGL, there's three main places where we have to worry about dma-buf feedback:
- In `eglGetPlatformDisplay`, we have to decide whether to return an EGLDisplay, or to return NULL and let Mesa handle the wl_display.
- In `eglInitialize`, we have to pick a device to use for rendering, and once we select it, we have to stick with it until `eglTerminate`.
- After the application creates an EGLSurface, we have to decide whether to render directly to the `wl_buffer`'s dma-buf, or if we need to render to a shadow buffer and blit to the `wl_buffer`.

In `eglGetPlatformDisplay`, we have to decide if we're going to claim the wl_display at all. Libglvnd will call each driver's `eglGetPlatformDisplay` implementation until the first one returns non-NULL.

This is where `__NV_PRIME_RENDER_OFFLOAD` comes into play. The default behavior (if `__NV_PRIME_RENDER_OFFLOAD` is *not* set) is that the application should end up rendering with whatever the system's "default" device is. For linux-dmabuf-v5, that means the main device, and for v6, that means the first sampling device.

If that "default" device is an NVIDIA device, then our `eglGetPlatformDisplay` should return non-NULL. If it's not an NVIDIA device, then we return NULL, so that Mesa can handle it instead.

If `__NV_PRIME_RENDER_OFFLOAD` *is* set, then that means the user wants the application to run on an NVIDIA device if at all possible, and so we should return non-NULL if we can present at all. That should be true if either of these holds:
- At least one tranche specifies an NVIDIA device, and has the sampling flag set.
- Or, at least one tranche (with any device, NVIDIA or not, sampling or not) accepts a pitch linear image.

Technically, in `eglGetPlatformDisplay`, we don't care about beyond those two points, because don't have to select a rendering device until `eglInitialze`. In practice, since we already have to assume that the feedback from `zwp_linux_dmabuf_v1::get_default_feedback` is static anyway, it doesn't really matter if we pick an EGLDeviceEXT in `eglGetPlatformDisplay` or defer that until `eglInitialize`. egl-wayland does the former, egl-wayland2 does the latter.

If we do defer the EGLDeviceEXT selection until `eglInitialize`, then `__NV_PRIME_RENDER_OFFLOAD` doesn't matter at that point. We've already claimed the EGLDisplay, so one way or another, we have find something that works.

Anyway, we have to pick an EGLDevice for rendering and we have to stick with it. For that, we'll probably want to pick a sampling device if there is one. If we don't find a sampling device, then we should probably pick the first NVIDIA device we find in the dma-buf feedback. If no tranche lists an NVIDIA device at all, then we can pick an arbitrary EGLDeviceEXT.

In addition to picking an EGLDeviceEXT, `eglInitialize` also needs to look at the format/modifier lists in the feedback data to figure out which EGLConfigs should have EGL_WINDOW_BIT set.

Note that all of the above has to work with the default feedback (i.e., `zwp_linux_dmabuf_v1::get_default_feedback`). We won't see a `wl_surface` until the first call to `eglCreateWindowSurface`, and you can't call that until after `eglInitialize` finishes.

That also means we're only going to look at the first round of feedback data, and then destroy the `zwp_linux_dmabuf_feedback_v1` object before `eglInitialize` returns. We can't switch to another device, and we can't change the EGLConfig attributes, so there's nothing we can do with any updates. We just have to assume that the original feedback data remains valid.

Finally, when the application actually creates an `EGLSurface`, that's when we can use `zwp_linux_dmabuf_v1::get_surface_feedback` to get per-surface feedback, and when we'll keep tabs on any updates to that feedback. Based on that feedback, we can switch between our two presentation modes on the fly: If the rendering device is also a sampling device at the moment, then we can render directly into the `wl_buffer`. If the rendering device is not a sampling device, then we'll need to render to a shadow buffer and blit.

amshafer:
Updated with some fixes after testing with cosmic, at this point this is working well. Also tested with kwin/mutter that this still works with dmabuf v5 compositors. Removing draft status. The build checks are failing due to ubuntu not having wayland-protocols with v6 in their lts releases, so we'll have to figure out what we want to do about that.

oscarbg:
sorry for asking here: but any word from NVIDIA on if next driver series R615 will have this support?.. or maybe for R620 branch series.. also if it will come perhaps earlier in Nvidia Vulkan developer driver or makes no sense?

src/wayland-egldisplay.c
```diff
@@ -560,8 +575,30 @@ static void
...
+    dev_t targetDev = protocols->currentTranche.targetDev;
+
+    /* Track the first sampling tranche regardless of GPU vendor */
+    if (protocols->first_sample_dev == 0) {
+        protocols->first_sample_dev = targetDev;
+    }
+
+    /* Check if this sampling tranche targets an NVIDIA GPU */
+    if (protocols->nvidia_sample_dev == 0 &&
+        protocols->pData->getDeviceFromDevId(targetDev, 0, &drm_device) == 0) {
```
kbrenneman: Note that this won't be enough to identify a Tegra device, since those aren't PCI devices. We have to use `drmGetVersion` to identify those.

kbrenneman:
I think the logic for whether `wlEglGetPlatformDisplayExport` should return NULL isn't quite correct.

If I'm reading it right, then `dmabuf_feedback_check_tranche_done` will set `WlServerProtocols::nvidia_sample_dev` if *any* tranche lists an NVIDIA device and has the sampling flag set. Then, `dmabuf_feedback_check_done` will set `WlServerProtocols::drm_name` to whatever node `nvidia_sample_dev` refers to, and `wlEglGetPlatformDisplayExport` will always try to use that node.

If `__NV_PRIME_RENDER_OFFLOAD` is not set, then we should only ever look at the first sampling device, because the first samling device is what should handle rendering by default. So, if that's not an NV device, then `wlEglGetPlatformDisplayExport` should return NULL and let Mesa handle the display.

If `__NV_PRIME_RENDER_OFFLOAD` *is* set, then we can look for an NVIDIA sampling device if one is available to decide which EGLDeviceEXT to use when we create the internal EGLDisplay.

src/wayland-egldisplay.c
```diff
@@ -572,6 +609,30 @@ dmabuf_feedback_check_done(void *data, struct zwp_linux_dmabuf_feedback_v1 *dmab
...
+     *   save this device for use as the set_sampling_device hint.
+     *
+     * Clear per-cycle state so the next feedback cycle starts fresh.
+     */
+    if (protocols->dmabufVersion >= 6) {
+        if (protocols->nvidia_sample_dev != 0) {
+            protocols->devId = protocols->nvidia_sample_dev;
+        } else if (protocols->first_sample_dev != 0) {
+            protocols->devId = protocols->first_sample_dev;
+            protocols->primeSamplingDev = protocols->first_sample_dev;
```
kbrenneman: Before we assume that we can use this for PRIME, we should make sure that it actually supports pitch linear modifiers.

That's something that we'd need to check in `dmabuf_feedback_check_tranche_done`, because we can use PRIME if *any* sampling device supports linear, not just first one.

But, we still need to keep track of the first sampling device, because if `__NV_PRIME_RENDER_OFFLOAD` isn't set, then that will determine whether or not `wlEglGetPlatformDisplayExport` should return NULL.

src/wayland-eglsurface.c
```diff
@@ -1380,6 +1382,27 @@ acquire_surface_image(WlEglDisplay *display, WlEglSurface *surface)
...
+         */
+        wl_array_init(&wlDev);
+        if (display->dmaBufProtocolVersion >= 6
+            && (wlDevData = (dev_t *)wl_array_add(&wlDev, sizeof(dev_t)))) {
+            if (surface->primeRenderOffload && surface->primeSamplingDevice != 0) {
+                *wlDevData = surface->primeSamplingDevice;
+            } else {
+                *wlDevData = display->devDpy->renderNode;
+            }
+            zwp_linux_buffer_params_v1_set_sampling_device(params, &wlDev);
```
kbrenneman, line 1398: Should we send the sampling device at all in the PRIME case? In the PRIME path, the shared dma-buf is in sysmem, so any device should be able to import it.

src/wayland-eglsurface.c
```diff
@@ -2266,6 +2296,51 @@ WlEglSurface *wlEglCreateSurfaceExport2(EGLDisplay dpy,
...
+    for (i = 0; i < feedback->numTranches; i++) {
+        WlEglDmaBufTranche *tranche = &feedback->tranches[i];
+
+        if (!tranche->supportsSampling)
+            continue;
+
+        if (first_sampling_dev == 0)
+            first_sampling_dev = tranche->drmDev;
+
+        if (tranche->drmDev == nvidia_dev) {
```
kbrenneman: I think technically the compositor would be allowed to send a primary node for one round of feedback and a render node for another, which would trip this up.

amshafer: Updated

src/wayland-eglsurface.c
amshafer: I don't think it hurts sending it, I agree it's not a hard requirement but it does hint to the compositor that we intend for this to be imported on the primary GPU. It's only a hint so I don't feel that strongly about it, and as you said the buffer will be importable anywhere if the compositor wants to do so.

src/wayland-egldisplay.c
amshafer: Updated

src/wayland-egldisplay.c
amshafer: We do check that in `checkNvidiaDrmDevice` already so I unified that to also be called from here

src/wayland-egldisplay.c
```diff
@@ -544,42 +567,143 @@ dmabuf_feedback_check_main_device(void *data,
...
 dmabuf_feedback_check_tranche_flags(void *data,
                               struct zwp_linux_dmabuf_feedback_v1 *dmabuf_feedback,
                               uint32_t flags)
 {
-    (void) data;
+    WlServerProtocols *protocols = (WlServerProtocols *)data;
     (void) dmabuf_feedback;
-    (void) flags;
+
+    protocols->currentTranche.sampled = flags & ZWP_LINUX_DMABUF_FEEDBACK_V1_TRANCHE_FLAGS_SAMPLING;
```
kbrenneman: This should be `(flags & ZWP_LINUX_DMABUF_FEEDBACK_V1_TRANCHE_FLAGS_SAMPLING) != 0`, so that you get a valid boolean value.

src/wayland-egldisplay.c
```diff
@@ -544,42 +567,143 @@ dmabuf_feedback_check_main_device(void *data,
...
+        return;
+    }
+
+    /* Record whether this tranche can import LINEAR buffers. */
+    wl_array_for_each(index, indices) {
+        if (*index >= protocols->formatTableLen) {
+            /* Out of range index, compositor bug. Skip it. */
+            continue;
+        }
+        if (protocols->formatTable[*index].modifier == DRM_FORMAT_MOD_LINEAR) {
```
kbrenneman: Should this also check the fourcc code? It's possible that a compositor might advertise linear support on some formats but not others.

Even if we're not correlating modifier support to specific EGLConfigs, looking for modifier support specifically for XRGB8 should work, since we can expect every compositor to support that format.

src/wayland-egldisplay.c
```diff
@@ -544,42 +567,143 @@ dmabuf_feedback_check_main_device(void *data,
...
-    (void) data;
+    WlServerProtocols *protocols = (WlServerProtocols *)data;
+    drmDevice *drm_device;
     (void) dmabuf_feedback;
+
+    if (protocols->currentTranche.sampled) {
+        dev_t targetDev = protocols->currentTranche.targetDev;
+
+        /* Track the first sampling tranche regardless of GPU vendor */
+        if (protocols->first_sample_dev == 0) {
```
kbrenneman, line 680: I'm not sure if how portable it is to assume that zero will never be a valid dev_t for a DRM device.

I'm *think* that's a safe assumption on Linux, but I don't know about elsewhere.

src/wayland-egldisplay.c
```diff
@@ -645,9 +821,11 @@ registry_handle_global_check_protocols(
     if ((strcmp(interface, "zwp_linux_dmabuf_v1") == 0) &&
         (version >= 3)) {
         protocols->hasDmaBuf = EGL_TRUE;
+        protocols->dmabufVersion = version;
```
kbrenneman: We should set this to whatever version we bind, rather than what the server advertises.

Or alternately, we could just remove `dmabufVersion` and call `wl_proxy_get_version` whenever we need to check the version.

src/wayland-egldisplay.c
amshafer: Seems this is safe on linux and freebsd, so I think that's enough? iirc we've used zero for this before.

Since we were talking about Rust recently, `Option` would be convenient here.

src/wayland-eglsurface.c
kbrenneman: I also realized that it actually *does* matter for PRIME, because whether or not we can use implicit sync depends on whether the sampling device is an NV device.

kbrenneman:
I *think* this looks right, but I'm still trying to reason through what would happen if we get per-surface feedback that's totally unusable (i.e., rendering GPU isn't sampling, and no sampling device that advertises linear support).

src/wayland-eglsurface.c
amshafer: What do you mean by "use implicit sync"? You mean use our background thread workaround path? Wouldn't that be affected by if we detected an NVIDIA sampling device, not by if we send the sampling device hint to the server?

amshafer:
In that case we would now fail to load the driver, which is probably fine.

Although the odds of that situation happening are extremely slim. Afaik all drivers for vendors used in optimus machines support the linear modifier so I anticipate we would always find the prime path usable. Vendors essentially *must* support the linear modifier if they want their hardware to be usable in a multi-vendor environment so I think it's reasonable to expect support for it.

src/wayland-eglsurface.c
kbrenneman: I meant using DMA_BUF_IOCTL_IMPORT/EXPORT_SYNC_FILE to attach the sync fd to the dma-buf for Mesa to use it in the PRIME path -- we'd be able to do that only if the server imported that dma-buf into a non-NV driver.

But from a quick grep, it looks like we never implemented that in egl-wayland anyway, so I guess it doesn't matter.

kbrenneman:
> Afaik all drivers for vendors used in optimus machines support the linear modifier so I anticipate we would always find the prime path usable.

Our driver is the notable exception to that, because a compositor would need to use GL_TEXTURE_EXTERNAL to import a linear dma-buf. If a compositor doesn't handle external-only modifiers, then that means we wouldn't be able to use an NV device (other than the rendering device) as a sampling device. And if that's the *only* sampling device that the compositor offers, then we'd end up with no usable modifiers.

We should definitely fail eglGetPlatformDisplay/eglInitialize if the default feedback gives us an unusable configuration, but it's technically possible for the original default feedback to be usable, but then get per-surface feedback that isn't.

amshafer:
Even if the compositor didn't handle external-only modifiers wouldn't we still be able to use the blocklinear modifiers that we share with the NVIDIA device? I don't think those are external-only even for cross-devices and both devices in this scenario would be NVIDIA?

kbrenneman:
We could use block linear, but we don't have any way to allocate a block linear buffer in sysmem. Linear works because our driver has a special case for `EGL_NV_stream_consumer_eglimage` that puts the linear buffer in sysmem.

Zamundaaa:
> no sampling device that advertises linear support

FWIW that case is already broken (we have a bug report for Nvidia as primary + Intel as secondary at least), so it wouldn't be a regression afaict.

> if the compositor didn't handle external-only modifiers

Unfortunately that's kind of required, because Xwayland still makes the assumption that all advertised modifiers (for 8bpc rgb formats) are usable for rendering. KWin *can* handle external_only just fine, but it can't advertise that to apps or Xwayland breaks.

kbrenneman:
> > no sampling device that advertises linear support
> 
> FWIW that case is already broken (we have a bug report for Nvidia as primary + Intel as secondary at least), so it wouldn't be a regression afaict.
> 
> > if the compositor didn't handle external-only modifiers
> 
> Unfortunately that's kind of required, because Xwayland still makes the assumption that all advertised modifiers (for 8bpc rgb formats) are usable for rendering. KWin _can_ handle external_only just fine, but it can't advertise that to apps or Xwayland breaks.

The fact that XWayland doesn't check the external-only flag is a bug in glamor, albeit a longstanding one. The best solution to that would be to just fix it.

kbrenneman:
Longer term, I'm hoping that we can get a direct peer-to-peer blit working for the NV -> NV case, rather than having to do a double blit from vidmem to sysmem to vidmem. By the time we have that, though, egl-wayland will probably have been fully retired in favor of egl-wayland2.

amshafer:
Sounds like the current approach in this MR is the best we can do then? Best-effort where we check that the initial configuration is valid and ensure there is linear support for the prime path? If the P2P path becomes available we can use it in the future

kbrenneman:
Yeah, I'd say that if the compositor changes its configuration on the fly such that per-surface feedback precludes something that default feedback allowed, then that's not something that an EGL client can reasonably deal with.

src/wayland-eglsurface.c
```diff
@@ -1787,11 +1810,18 @@ static EGLint create_surface_stream_local(WlEglSurface *surface)
...
+             * in a prime setup, look up the sampling device's format set so we
+             * can check which modifiers it supports (typically linear-only).
+             *
+             * On dmabuf v6+ the sampling device comes from the per-surface
+             * feedback tranches (set by wlEglUpdateSurfacePrimeState).  On
+             * earlier versions the main_device from the feedback is used.
              */
-            if (!formatSet && display->primeRenderOffload) {
-                formatSet = WlEglGetFormatSetForDev(feedback, feedback->mainDev, format);
+            if (!formatSet && surface->primeRenderOffload) {
```
kbrenneman: If we're going through PRIME for presentation, then we have to use pitch linear, because our driver special cases pitch linear to put the dma-buf in sysmem.

If the rendering and sampling device are two different NV GPUs, then this might find a common block linear modifier. If we tried to use a block linear modifier, then it would get allocated in vidmem, and the sampling device wouldn't be able to access it.

amshafer: This did exist in previous versions but based on above discussion makes sense to change, I updated it to force the prime path to select linear.

src/wayland-eglsurface.c
```diff
@@ -1805,6 +1820,19 @@ static EGLint create_surface_stream_local(WlEglSurface *surface)
...
         }
+
+        /*
+         * PRIME render offload copies buffers between the rendering and
+         * sampling devices, which only reliably works with linearly laid
+         * out buffers. Force the linear modifier here instead of trusting
+         * the (possibly stale or overly permissive) advertised modifier
+         * list.
+         */
+        if (surface->primeRenderOffload) {
```
kbrenneman, line 1831: Since we're overwriting the modifier array, I think we could skip the whole previous if/else block when `primeRenderOffload` is set.

src/wayland-eglsurface.c
```diff
@@ -1744,6 +1767,7 @@ static EGLint create_surface_stream_local(WlEglSurface *surface)
     EGLint err = EGL_SUCCESS;
     EGLint numModifiers = 0;
     EGLuint64KHR *modifiers = NULL;
+    EGLuint64KHR linearModifier;
```
kbrenneman, line 1770: Minor nit, but this could be a static const, rather than needing to assign it below.

amshafer: Made it static, couldn't make it const since that would mismatch the pointer usage and throw a qualifier warning.

kbrenneman: Wait, which usage?

`eglStreamImageConsumerConnect` takes a const pointer for the `modifiers` parameter, and we don't modify it that function, so we should be able to make the local `modifiers` variable const.

amshafer: Ah sorry, yea it was the local modifiers variable. Updated

kbrenneman (approved):
Looks good to me.

elieltondias (approved):

PR #190, opened 2026-03-23
egl-wayland: Prefer render node over primary node

Molytho:
This fixes an issue where the user has permissions for the render node but not the primary node.

This is an issue I encountered in a headless setup where the render node is readable and writeable by every user while the primary node is only accessible by root (because its permissions are managed by systemd-logind in a non-headless setup)

### DIFF
--- a/src/wayland-egldisplay.c
+++ b/src/wayland-egldisplay.c
@@ -949,6 +949,7 @@ EGLDisplay wlEglGetPlatformDisplayExport(void *data,
     EGLBoolean usePrimeRenderOffload = EGL_FALSE;
     EGLBoolean isServerNV;
     const char *drmName = NULL;
+    const char *devExts = NULL;
 
     if (platform != EGL_PLATFORM_WAYLAND_EXT) {
         wlEglSetError(data, EGL_BAD_PARAMETER);
@@ -1191,13 +1192,24 @@ EGLDisplay wlEglGetPlatformDisplayExport(void *data,
     WL_LIST_INIT(&display->wlEglSurfaceList);
 
     /* Get the DRM device in use */
-    drmName = display->data->egl.queryDeviceString(display->devDpy->eglDevice,
-                                                   EGL_DRM_DEVICE_FILE_EXT);
-    if (!drmName) {
+    devExts = display->data->egl.queryDeviceString(display->devDpy->eglDevice, EGL_EXTENSIONS);
+    if (!devExts) {
         goto fail;
     }
 
-    display->drmFd = open(drmName, O_RDWR | O_CLOEXEC);
+    display->drmFd = -1;
+    if (wlEglFindExtension("EGL_EXT_device_drm_render_node", devExts)) {
+        drmName = display->data->egl.queryDeviceString(display->devDpy->eglDevice, EGL_DRM_RENDER_NODE_FILE_EXT);
+        if (drmName) {
+            display->drmFd = open(drmName, O_RDWR | O_CLOEXEC);
+        }
+    }
+    if (display->drmFd < 0 && wlEglFindExtension("EGL_EXT_device_drm", devExts)) {
+        drmName = display->data->egl.queryDeviceString(display->devDpy->eglDevice, EGL_DRM_DEVICE_FILE_EXT);
+        if (drmName) {
+            display->drmFd = open(drmName, O_RDWR | O_CLOEXEC);
+        }
+    }
     if (display->drmFd < 0) {
         goto fail;
     }


src/wayland-egldisplay.c
```diff
@@ -929,6 +929,18 @@ static EGLBoolean checkNvidiaDrmDevice(WlServerProtocols *protocols)
     return result;
 }
 
+static const char *getDrmDeviceNode(WlEglPlatformData *platform_data, EGLDeviceEXT device, bool render_node_only) {
+    const char *dev_exts = platform_data->egl.queryDeviceString(device, EGL_EXTENSIONS);
+    if (dev_exts && wlEglFindExtension("EGL_EXT_device_drm_render_node", dev_exts)) {
+        return platform_data->egl.queryDeviceString(device, EGL_DRM_RENDER_NODE_FILE_EXT);
+    } else if (render_node_only) {
```
kbrenneman: If an EGLDeviceEXT doesn't advertise `EGL_EXT_device_drm_render_node`, then `EGL_DRM_DEVICE_FILE_EXT` is allowed to be the primary or render node. In particular, that can happen if a device doesn't have a primary node at all.

src/wayland-egldisplay.c
```diff
@@ -1191,8 +1196,7 @@ EGLDisplay wlEglGetPlatformDisplayExport(void *data,
     WL_LIST_INIT(&display->wlEglSurfaceList);
 
     /* Get the DRM device in use */
-    drmName = display->data->egl.queryDeviceString(display->devDpy->eglDevice,
-                                                   EGL_DRM_DEVICE_FILE_EXT);
+    drmName = getDrmDeviceNode(display->data, display->devDpy->eglDevice, false);
```
kbrenneman: It would probably be better to try both nodes here: Try opening the render node first (if EGL_EXT_device_drm_render_node is there), and if that fails, try the primary node.

Molytho: I pushed a version that does this

kbrenneman: What you've got now is closer, but the code duplication and redundant fallback paths make it harder to read than it needs to be.

As currently structured, the initialization code needs to do two things with EGLDevices:
1. Given a device node path, find the corresponding EGLDeviceEXT
2. Given an EGLDeviceEXT, open a file descriptor to that device (which can be the primary or render node)

For (1), the existing code is probably fine as-is, unless there's already some compositor out there which sends the primary node back using wl_drm instead of the render node. Even then, trying to factor out the `eglQueryDeviceStringEXT` calls doesn't reduce duplication, and makes the logic much harder to follow.

For (2), the only change that's needed is that if `open()` fails on one node, then it should look up the other node and try again. So all you really need is something like:
```
int fd = -1;
if (EGL_EXT_device_drm_render_node is supported) {
    const char *node = eglQueryDeviceStringEXT(egldev, EGL_DRM_RENDER_NODE_FILE_EXT);
    fd = open(node, O_RDWR);
}
if (fd < 0) {
    const char *node = eglQueryDeviceStringEXT(egldev, EGL_DRM_DEVICE_FILE_EXT);
    fd = open(node, O_RDWR);
}
if (fd < 0) {
    goto fail;
}
```

With the appropriate checks for NULL, of course.

Molytho: I didn't like the code I wrote either. It looks way better now.

kbrenneman:
The new logic in this looks right, I think.

The extension string check in `supportsDrmRenderNodeEXT` is short and simple enough that it doesn't need to be factored out, though. Just calling `eglQueryDeviceString` and `wlEglFindExtension` from within `wlEglGetPlatformDisplayExport` is fine and easier to read.

src/wayland-egldisplay.c
```diff
@@ -1191,13 +1194,19 @@ EGLDisplay wlEglGetPlatformDisplayExport(void *data,
...
+        drmName = display->data->egl.queryDeviceString(display->devDpy->eglDevice, EGL_DRM_RENDER_NODE_FILE_EXT);
+        if (drmName) {
+            display->drmFd = open(drmName, O_RDWR | O_CLOEXEC);
+        }
+    }
+    if (display->drmFd < 0) {
+        drmName = display->data->egl.queryDeviceString(display->devDpy->eglDevice, EGL_DRM_DEVICE_FILE_EXT);
+        if (drmName) {
+            display->drmFd = open(drmName, O_RDWR | O_CLOEXEC);
+        }
```
amshafer, line 1211: I think this change seems reasonable.

One preexisting thing though is do we also need to check if the `EXT_device_drm` extension is supported? So far I think we have inferred that it has to be supported since we know DRM is in use by the compositor, but we don't actually check for it anywhere. Maybe we want to officially check it here while we are at it with a similar `supportsDeviceDrmEXT`?

kbrenneman: It wouldn't hurt to add that check at some point, though there's no need to hold up this PR for it.

That hasn't been an issue so far because every system where we support platform libraries (except QNX, maybe?) also always supports EGL_EXT_device_drm. And if EGL_EXT_device_drm was missing, then it just means that eglQueryDeviceStringEXT would return NULL, which we already handle.

amshafer: Yea fair enough, I don't want to hold this up

amshafer (approved):

Molytho:
I removed the supportsDrmRenderNodeEXT function and while I was at it I also added the check for EGL_EXT_device_drm

ehfd:
@amshafer @kbrenneman Hi, any status of this PR?

PR #195, opened 2026-07-07
Fix fd leaks in linux-dmabuf feedback handling

qdddddd:
## Problem

Long-lived EGL clients leak one file descriptor per `zwp_linux_dmabuf_feedback_v1::format_table` event that the compositor resends, until the process hits `RLIMIT_NOFILE`. At that point the kernel silently drops the next `SCM_RIGHTS` fd (see `unix(7)`), libwayland fails demarshalling with:

```
file descriptor expected, object (N), message format_table(hu)
[glfw error 65544]: Wayland: fatal display error: Invalid argument
```

and the client exits. Observed in the wild with kitty 0.47.4 on GNOME 46 (mutter 46.2) with NVIDIA 580.159.03: the client process accumulated hundreds of open 4032-byte `/memfd:mutter-shared` fds (each exactly the 252-entry format+modifier table) at ~24–38 fds/min under frequent feedback resends, and crashed roughly every 40 minutes against the default 1024 soft limit.

This is the client-side half of a known cross-stack issue:

- https://gitlab.freedesktop.org/wayland/wayland/-/issues/594 (closed as not-a-libwayland-bug: clients must not exhaust fds; connection death is unrecoverable by design)
- https://gitlab.gnome.org/GNOME/mutter/-/issues/4573, fixed compositor-side by https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/4838 — but only in mutter >= 50; mutter 46/48/49 deployments (e.g. Ubuntu 24.04 LTS) resend the format table with a fresh fd on every scanout-candidacy change and remain in the field for years
- https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/4088 (same compositor-side issue in wlroots, fixed May 2026)

## Root causes in egl-wayland

1. **The default feedback proxy outlives the last dispatch of its queue.** The default dmabuf feedback is bound on the display's internal `wlEventQueue`, which is never dispatched again after `eglInitialize` completes. Every feedback resend therefore parks a closure holding an open format-table fd on that queue forever. The parsed data (format table, tranches) lives in `defaultFeedback` and remains valid; resent updates were never processed anyway. This PR destroys the proxy once the initial burst has been handled — per-surface feedback still tracks dynamic changes on surface queues, which are dispatched at swap time. (If maintainers prefer keeping the proxy alive and instead dispatching the queue periodically, happy to rework.)

2. **`dmabuf_feedback_check_format_table()` ignores the fd without closing it** — one leaked fd per EGL display initialization since the probe was introduced in 1.1.14.

3. **`dmabuf_feedback_format_table()` never unmaps the previous table on a resend**, leaking the old mapping each time a resend is actually dispatched (per-surface feedback path).

## Testing

- Built on Ubuntu 24.04 (driver 580.159.03, mutter 46.2, wayland 1.22) and run under kitty: rendering, resize, and fullscreen toggling behave normally.
- `WAYLAND_DEBUG=1` confirms the default feedback proxy is now destroyed right after the init roundtrip (`zwp_linux_dmabuf_feedback_v1.destroy()` immediately follows the initial `done()`), so compositors stop resending to it entirely.
- Leak measurable before/after via `ls -l /proc/<pid>/fd | grep -c memfd` under a compositor that resends feedback.

### DIFF
--- a/include/wayland-egldisplay.h
+++ b/include/wayland-egldisplay.h
@@ -129,7 +129,6 @@ typedef struct WlEglDisplayRec {
     EGLBoolean         ownNativeDpy;
     struct wl_display *nativeDpy;
 
-    struct wl_registry             *wlRegistry;
     struct wl_eglstream_display    *wlStreamDpy;
     struct wl_eglstream_controller *wlStreamCtl;
     struct zwp_linux_dmabuf_v1     *wlDmaBuf;
--- a/src/wayland-egldisplay.c
+++ b/src/wayland-egldisplay.c
@@ -743,10 +743,6 @@ static EGLBoolean terminateDisplay(WlEglDisplay *display, EGLBoolean globalTeard
         wlEglDestroyFormatSet(&display->formatSet);
         wlEglDestroyFeedback(&display->defaultFeedback);
 
-        if (display->wlRegistry) {
-            wl_registry_destroy(display->wlRegistry);
-            display->wlRegistry = NULL;
-        }
         if (display->wlStreamDpy) {
             wl_eglstream_display_destroy(display->wlStreamDpy);
             display->wlStreamDpy = NULL;
@@ -1325,6 +1321,7 @@ EGLBoolean wlEglInitializeHook(EGLDisplay dpy, EGLint *major, EGLint *minor)
     WlEglDisplay      *display = wlEglAcquireDisplay(dpy);
     WlEglPlatformData *data    = NULL;
     struct wl_display *wrapper = NULL;
+    struct wl_registry *wlRegistry = NULL;
     EGLint             err     = EGL_SUCCESS;
     int                ret     = 0;
     const char *dev_exts = NULL;
@@ -1388,9 +1385,9 @@ EGLBoolean wlEglInitializeHook(EGLDisplay dpy, EGLint *major, EGLint *minor)
     /* Listen to wl_registry events and make a roundtrip in order to find the
      * wl_eglstream_display and/or zwp_linux_dmabuf_v1 global object
      */
-    display->wlRegistry = wl_display_get_registry(wrapper);
+    wlRegistry = wl_display_get_registry(wrapper);
     wl_proxy_wrapper_destroy(wrapper); /* Done with wrapper */
-    ret = wl_registry_add_listener(display->wlRegistry,
+    ret = wl_registry_add_listener(wlRegistry,
                                    &registry_listener,
                                    display);
     if (ret == 0) {
@@ -1462,9 +1459,9 @@ EGLBoolean wlEglInitializeHook(EGLDisplay dpy, EGLint *major, EGLint *minor)
         display->defaultFeedback.wlDmaBufFeedback = NULL;
     }
 
-    if (display->wlRegistry) {
-        wl_registry_destroy(display->wlRegistry);
-        display->wlRegistry = NULL;
+    if (wlRegistry) {
+        wl_registry_destroy(wlRegistry);
+        wlRegistry = NULL;
     }
 
     if (major != NULL) {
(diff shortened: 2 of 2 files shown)


amshafer:
> If maintainers prefer keeping the proxy alive and instead dispatching the queue periodically, happy to rework

I would definitely prefer this approach, please update the fix to implement it. We definitely want to update the default feedback when the compositor sends a new version.

The other two fixes seem reasonable to me. cc @kbrenneman to take a second look.

kbrenneman:
I think destroying the default feedback proxy is the correct behavior. To start with, there's no good time to dispatch the display-level queue.

But even if we did get new default feedback, there's nothing we can do with it: After `eglInitialize` returns, we can't change the rendering device, and we can't change the EGLConfig list.

For updating EGLSurfaces, we use per-surface feedback, and each EGLSurface gets its own `wl_event_queue` to receive those.

Note that egl-wayland2 not only destroys the default `zwp_linux_dmabuf_feedback_v1` proxy, it doesn't even keep the `wl_event_queue` around after `eglInitialize` finishes.

amshafer:
Eh fair enough, it would probably also be good to have both egl-wayland versions matching in behavior as much as possible

Drakulix:
> But even if we did get new default feedback, there's nothing we can do with it: After `eglInitialize` returns, we can't change the rendering device, and we can't change the EGLConfig list.

You can with https://github.com/NVIDIA/egl-wayland/pull/178, just in case that is a consideration for adding some preparations in this PR.

qdddddd:
Thanks for the review. Following the discussion, I've kept the destroy-after-init approach and pushed a commit aligning this further with egl-wayland2: the display registry is now also destroyed once initialization completes, since it's only used for global discovery during init and runtime global announcements would otherwise accumulate on the same never-dispatched queue.

I stopped short of dropping the `wl_event_queue` itself as egl-wayland2 does: in this codebase `wlDmaBuf`, `wpPresentation`, and friends stay attached to it for the display's lifetime, and the EGLStream path still does swap-interval roundtrips on it (`wayland-eglswap.c:86,272`). With the feedback proxy and registry gone, nothing can accumulate on the queue on the dmabuf path anyway. Buffer creation is unaffected — params are created through a wrapper re-queued to the surface's buffer queue and use `create_immed`.

@Drakulix regarding #178: this PR only destroys the feedback *proxy* — `display->wlDmaBuf` stays alive, so the v6 code can call `get_default_feedback` again whenever it needs live updates (e.g. reacting to main-device changes to drive `set_sampling_device`), ideally bound to a queue it dispatches deliberately. Re-creating the proxy on demand seemed cleaner than keeping one parked on a queue that is never dispatched, so I don't believe this change precludes anything #178 needs.

Retested on GNOME 46 / NVIDIA 580.159.03: rendering, resize, and fullscreen toggling behave normally with all three commits applied.

amshafer:
LGTM, we can land this once @kbrenneman signs off

kbrenneman:
Given that the `wl_registry` doesn't outlast `wlEglInitializeHook`, we don't need `WlEglDisplay::wlRegistry` anymore -- that should just be a local variable instead.

We should also munmap the format table when we destroy the default feedback proxy, since we won't need it after that.

qdddddd:
Done, pushed three commits addressing both points:

- `wlRegistry` is now a local variable in `wlEglInitializeHook`. It is also destroyed on the failure path, since `terminateDisplay()` no longer handles it.
- The format table is unmapped when the default feedback proxy is destroyed. The unmap moved into a shared helper (`wlEglFeedbackUnmapFormatTable`) so the resend and teardown paths use the same guards; it clears `entry`/`len`, keeping the later `wlEglDestroyFeedback` at terminate a no-op for the mapping.
- While in there: `wlEglDestroyFeedback` now also clears `wlDmaBufFeedback` after destroying it, so a repeated call on the same feedback (e.g. teardown after a failed `eglInitialize` retry) cannot destroy a stale proxy.

kbrenneman:
That all looks correct, I think.

kbrenneman (approved):

qdddddd:
Just a reminder that the workflows are pending approval

amshafer:
Seems reasonable to me, thanks

